PR #3, opened 2016-01-02
Redundant if condition removed

NJannasch:
Condition already checked in line 1009


tbeu:
Thanks for your three patches! HNY!

PR #7, opened 2016-01-06
Remove unused variable

NJannasch:
MAT_T_STRING is nowhere used
To store strings use MAT_T_UTF8


tbeu:
You are right about it. But since matio.h defines the official API to the library I won't delete it in order to not break any backward compatibility.

PR #8, opened 2016-01-06
Remove unused variable

NJannasch:
MAT_T_ARRAY is nowhere used, too


tbeu:
You are right about it. But since matio.h defines the official API to the library I won't delete it in order to not break any backward compatibility.

Thanks for the patches anyway.

NJannasch:
To my mind this change would not have a negative effect on the api because when you're upgrade your lib to a new version your code will fail anyway (if you have used the old variables). When the variables are not available it's easier to find the bug, too.

tbeu:
I am afraid I won't touch them even though they are dummies only.

NJannasch:
I can live with that ;-)
Thanks for your answer!

tbeu:
Indeed, no big deal at all.

PR #11, opened 2016-01-13
Fixes suggested by Coverity analysis (try 2)

gsjaardema:
Several fixes suggested by the Coverity analysis tool.  

Some of these are rather pedantic (such as prefixing functions with (void) to avoid warnings about ignoring return values); but some of the other changes do result in better error detection and reporting and do fix potential issues in the code.  

The coverity results should be able to be seen at https://scan.coverity.com/projects/matio?tab=overview


tbeu:
Thanks for the patches (just a few days after v1.5.3 was released thus it won't go in there anymore).

Why did you introduce Mat_Initialize()? It is nothing special about it and should be replaced by a simple call to calloc() in my opinion.

tbeu:
Adding (void) casts to fread does not help for GCC 4.8, see https://drone.io/github.com/tbeu/matio/173
```
mat5.c:5261:13: warning: ignoring return value of ‘fread’, declared with attribute warn_unused_result [-Wunused-result]
             (void)fread(tag,4,2,mat->fp);
             ^
```

gsjaardema:
Currently, Mat_Iniitialize could be replaced by calloc since all fields in
the struct are set to NULL or 0, but I guess I was preparing for a
potential use in the future where one or more of the fields would have a
non-zero value.  I guess I'm also primarily a C++ developer and this looked
more like a c++ constructur...

..Greg

tbeu:
* snprintf.c was completely updated such that your PR now conflicts with current HEAD.
 * The (void)fread is of no use for GCC.
 * The indent style should be blanks and not tabs.

Can you please provide single fixes (instead of one large commit) such that I can consider your fixes again. Thanks y lot for your collaboration.

PR #12, opened 2016-01-13
Let autogen.sh work on Darwin and Linux

gsjaardema:
Darwin/Mac systems do not have libtoolize (or they do and it has a different function), it is called glibtoolize.  The stackoverflow item http://stackoverflow.com/questions/15448582/installed-libtool-but-libtoolize-not-found has more information.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,7 +1,8 @@
 #! /bin/sh
 
 bootstrap() {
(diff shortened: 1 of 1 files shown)


tbeu:
I guess I need to setup a travis.ci environment to also test it on a mac.

PR #19, opened 2016-01-24
Add missing abort to Mat_Error routine

gsjaardema:
Description of routine states that it logs an error and then aborts, but it was missing the call to abort()

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -301,6 +301,7 @@ void Mat_Error( const char *format, ... )
     va_start(ap, format );
     mat_log( LOG_LEVEL_ERROR, format, ap );
     va_end(ap);
+    abort();
 }
 
 /** @brief Prints a helpstring to stdout and exits with status 1


tbeu:
Yes, Mat_Critical actually should not return to the calling function. Since this might quit any application (like Scilab or Freemat) using matio I'll leave it to the library users to implement Mat_Critical correctly. For matio as part of the Modelica Standard Library we call ModelicaError (from Mat_Critical), which must not return to the calling function by design, see https://github.com/tbeu/ModelicaStandardTables/blob/master/C-Sources/ModelicaMatIO.c#L1605-L1611 and https://github.com/tbeu/ModelicaStandardTables/blob/master/C-Sources/ModelicaUtilities.h#L94-L99. It is up to the tool vendor how to implement ModelicaError = Mat_Critical, but usually it is done by setjmp/longjmp, see https://trac.modelica.org/Modelica/ticket/1197.

gsjaardema:
But, if you leave it as it is and Mat_Critical does return by default, there will be undefined or at best a core dump.  It is fine if an application wants to override the behavior, but to have the default behavior be incorrect does not make sense to me.

gsjaardema:
Acutally, the routine being considered is Mat_Error and not Mat_Critical.  Not sure if that makes any difference in your response.

tbeu:
Sorry, I mixed it up: Mat_Error should not return to the calling function, whereas Mat_Critical should return to the calling function.

My above reasoning still holds.

tbeu:
Mat_Error is currently only used once in matio. I am not sure which of the other calls to Mat_Critical should rather call Mat_Error.

gsjaardema:
OK, in that case, can you add a README or some other file which states
which routines are buggy in matio and need to be overridden by the library
client?  In other words, how does a new client of the matio library know
that they need to provide a correct implementation of Mat_Error and are
there other routines where they need to do the same thing.

Maybe I am misunderstanding your reasoning, but I don't see why we want to
have a routine that is supposed to catch bad memory allocation or some
other fatal error and it prints out some error message and returns so the
code can then segfault or gives undefined behavior.

I think there are some memory allocation failures that are currently
calling Mat_Critical that should call Mat_Error and there are some memory
allocation failures that just fail.   I have also added Mat_Error calls to
several of the file positioning calls if they return errors.

If Mat_Error is not going to abort, then there needs to be some code
refactoring to make sure that an appropriate error code is set and returned
if the code should not continue the current routine after the error.  It
should call Mat_Error or Mat_Critical and then return to the calling
application with the correct error flag or code.

gsjaardema:
Note that I agree with you that a library error should not  abort a running
application.  The library should return an appropriate value to the calling
application and let the application decide how to proceed following the
error.  However, the code is currently not structured such that this is
possible in many cases -- there are memory allocation failures and the code
prints an error and then accesses the invalid memory or the code has a file
positioning error, prints an error and then accesses the file.

As an application developer, I would rather have the library print an error
and then fail "gracefully" then to print an error and then give me
undefined or arbitrary behavior.

Long term I would like to have the library print the error and then return
to my application with an error that my application can decide what to do.
Matio is not at that point yet.

..Greg

tbeu:
I talked about the error handling to Chris Hulbert some time ago. He simply said that he does not like it and that he would not do the error reporting and handling again like he did.

Maybe we should add an error handler that an application can provide. See the GSL Error handling for example: https://www.gnu.org/software/gsl/manual/html_node/Error-Handling.html

tbeu:
Then we can deal with Mat_Error (incl. abort()) as the default error handler. See again https://www.gnu.org/software/gsl/manual/html_node/Error-Handlers.html#Error-Handlers

tbeu:
Ah, actually there is already a default error handler and function. An application should call Mat_LogInitFunc with a customized log_func. The default error function matio_error_func already calls abort if the log_level is LOG_LEVEL_ERROR. Thus there is no need to change Mat_Error.

gsjaardema:
OK, I see that now.  Thanks for the explanation.   I may clean up the
commenting of the Mat_Error function to make this clearer to others (and
me) in the future.

tbeu:
Right. Thanks.

tbeu:
The noreturn attribute was added to function Mat_Error by 6fe8b0e7762237ba400c6351e721c3d532eaa1e9.

tbeu:
This was now added. Also I made sure that Mat_Error is not called from the matio lib itself.

PR #21, opened 2016-01-25
Update comments to make behavior clearer

gsjaardema:
Added comment to Mat_Error that mat_log calls abort and does not return.

Fix Mat_Help comment to indicate correct return value.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -299,13 +299,13 @@ void Mat_Error( const char *format, ... )
     va_list ap;
 
     va_start(ap, format );
-    mat_log( LOG_LEVEL_ERROR, format, ap );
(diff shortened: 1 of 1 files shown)


tbeu:
Function mat_log can be provided by the user through Mat_LogInitFunc in which case it is implementation dependent and your added comment /* Does not return, calls abort() */ may no longer hold in general. I'd propose to modify it to /* Shall never return to the calling function */ and provide some new description to function mat_log (which currently is not document at all).

tbeu:
Function mat_log can be provided by the user through Mat_LogInitFunc in which case it is implementation dependent and your added comment `/* Does not return, calls abort() */` may no longer hold in general. I propose to change it to `/* Shall never return to the calling function */` and some new description to mat_log which currently is not documented at all.

tbeu:
Merged.

PR #27, opened 2016-03-15
Minor Corrections

papadop:
This pull request corrects two small things.
Out of source build is broken because matio.sym is not found at the proper place (first commit).
The outputs of :
test_mat readvar matio_test_cases_hdf_le.mat  var23
test_mat readvar matio_test_cases_uncompressed_le.mat  var23
differ because the hdf reader sets matvar->data to NULL whereas it is non null in the second case (second commit).


papadop:
Oups. Forgot to correct the tests. Wait a minute.

papadop:
Yes our messages crossed. It is actually how I found the discrepancy. I'm trying to put the tests results in separate files.

papadop:
The last commit introduces the testsuite changes.

tbeu:
Thanks. All cherry picked manually.

PR #28, opened 2016-03-15
Tests reorganization

papadop:
This is not a complete pull request. It is just a preview to let you know what I intend to do. The idea is to remove redundancy so that tests results are stored once in a file and reused as much as possible across various tests. That's how I found the discrepancy in Mat_VarPrint between various matlab files this morning. It also somewhat simplifies maintenance. I probably need to check how to solve the line end problem on windows (I already solved it in an another testing framework).

It also allows an easier sharing of tests.

What do you think of the idea ? If you are opposed to it, then I guess I will not pursue it (but will also have no choice and will need to keep my fork). The long term goal is to merge as much as possible my fork with your repo (there might be things that will not be accepted in your repo in the end, but if we can minimize them, that will be best for all of us).


papadop:
Oups, it contains my previous changes. Obviously wrong... Please only consider commit 4ef9dbe here. As this is not the final pull request, I will cancel it once the discussion is over and will recreate a new one.

tbeu:
Yes, it would really be of help. There is lots of duplication in the expected test results w.r.t.
* little / big endian
* different mat file versions (uncompressed, compressed and hdf5)

If you can provide a sound and clean PR I will accept it for sure.

I run the tests on Windows using Cygwin, buth w/oz HDF5 support which I did not installed/built.

papadop:
Cool. I will continue then. But I will need to have the other pull request processed (accepted or not) so as to not re-modify the tests twice (for now I assume it will be accepted).

tbeu:
Please always commit your changes to individual branches which makes life easier here.

tbeu:
As for Windows, I only managed to succeed the testsuite in MSYS with MinGW if the diff tool was called with option `--strip-trailing-cr`. Do you have an idea how this diff option can be activated such that it is utilized at at_diff?

Just noticed that the diff option ``--ignore-trailing-space` might also be useful here.

tbeu:
Please note that all result files also need to go to the distribution.

papadop:
Well git can extract text files with their native line endings... So, normally, it should work almost naturally if git is properly configured...

tbeu:
It's longer git then once you call `make dist-gzip` to build the gzipped distribution.

PR #30, opened 2016-03-17
Use a C++ friendly va_copy detection.

papadop:
This is a very minor change that allows the compilation of matio with g++ (the va_copy macro detection fails because the exit function is not available by default in C++).


tbeu:
Thank you. Will merge it soon.

@tbeu: Add new file matio_va_copy.m4 to the distribution.

PR #31, opened 2016-03-27
Split matlab tests in their own files and share them across tests.

papadop:
This is the next step in test reorganisation. Matlab file are put in separate files and shared across the various tests. This also corrected a few tests that were blocking (error in the matlab file) at least on linux (mainly write_char tests). All tests now pass.

Note that I probably broke matlab tests previously by removing the newline in the PASSED chracter chain in the result. I think the testing framework is a little fragile here (due to the horrible syntax where [] are string delimiter), but I did not find any other way of doing things.

Quite a few matlab test files are parameterized by a type, to commonize tests differing by just a type.


tbeu:
Can you please attach the testsuite.log to this ticket. Thanks.

papadop:
File renamed testsuite.txt, otherwise github will not accept it.... But here it is....
[testsuite.txt](https://github.com/tbeu/matio/files/193679/testsuite.txt)

papadop:
New testsuite.log after last change (no matlab tests in that one).
[testsuite.txt](https://github.com/tbeu/matio/files/194547/testsuite.txt)

papadop:
With the last commit, I believe that now all tests results are split out of the .at files.

tbeu:
Do you also have the testsuite.log with MATLAB tests run?

papadop:
Well, I have not touched to matlab stuff with the last patches, so the previous testsuite.log should remains relevant here. But I will do that once at work.

papadop:
OK this work is now really finished (there were some more tests). There are some discrepancies between mat5 and mat73 tests results, but that will be dealt with in another branch. Overall, the testing framework is now 48.1% of what it was in size, and in my opinion it is now easier to find differences with the various matlab file formats. In a few minutes, I will post the full testsuite.log file (just waiting for it to finish on a clean tree).

tbeu:
Thanks. by coincidence I am just working on the merge right now...

papadop:
And the testsuite.log is:
[testsuite.txt](https://github.com/tbeu/matio/files/200247/testsuite.txt)

tbeu:
Nice to see them all succeeding. And takes less than 20min.

PR #32, opened 2016-03-27
C++

papadop:
This is another "just for discussion" pull request, even though it is already somewhat complete...
In OpenMEEG, we had complains of "strange" error messages and crashes when allocating big matrices. The problem is that OpenMEEG being a C++ program, we rely on exceptions to capture errors and fail gracefully. But matio uses plain malloc and justs exits (sometimes) when a memory allocation failure arises. In this branch, I replaced all memory allocations by macros that expand (or are equivalent to) the C code when compiled with C and to C++ allocation routines when compiling with C++. This is tested to work fine in both C and C++ compilations.... So now, if compiling in C++ mode, allocation failure now result in exceptions emitted. 

The drawback is that in C++ mode, no cleanup is made (because the if (var==NULL) blocks are not executed. To deal with that, I need to uglify  the code with macros simulating TRY {} CATCH() {}, which will become try {} catch(...) { CLEANUP } blocks in C++ mode and traditional if (var==NULL) {CLEANUP} blocks in C mode. 

One nice side effect of this work is that it is easy to identify all allocation points by grepping NEW. And if the TRY {}CATCH {} work is done, then all memory allocation failures will be caught, whereas this seems not to be the case currently.

Among other C++ improvements I might be interested to create are 
- to use the C++ IO system (this is because in OpenMEEG we want to read  files that are already open). But this requires an extension of the matio interface.
- wrap the various mat{4,5,7.3,,} routines in some C++ classes for easier use in C++ code. Of course, this has to be done without breaking the C interface
- extend the build system to build both a matio and a matio++ libraries.

Of course the first and second points need to be done without breaking the C interface. I have ideas, but I'm not totally sure if this can be done without uglifying the code too much. The last point is fairly easy (at least with the cmake build system I use, which will be the subject of another "just for discussion" pull request I will make once all the test reorganisation stuff will be integrated. With autoconf, this is doable but may need two separate builds (as it works currently, but we need to detect the C++ compiler and change the final library name).

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -163,7 +163,7 @@ mat_log(int loglevel, const char *format, va_list ap)
     if ( !logfunc ) return;
     buffer = strdup_vprintf(format, ap);
     (*logfunc)(loglevel,buffer);
-    free(buffer);
+    DELETE_ARRAY(buffer);
     return;
 }
 
--- a/src/io.c
+++ b/src/io.c
@@ -70,7 +70,7 @@ strdup_vprintf(const char* format, va_list ap)
     size = mat_vsnprintf(NULL, 0, format, ap2)+1;
     va_end(ap2);
 
-    buffer = (char*)malloc(size+1);
+    buffer = NEW_ARRAY(char,size+1);
     if ( !buffer )
         return NULL;
 
--- a/src/mat.c
+++ b/src/mat.c
@@ -322,7 +322,7 @@ Mat_Open(const char *matname,int mode)
     if ( NULL == mat )
         return mat;
 
-    mat->filename = strdup_printf("%s",matname);
+    mat->filename = STRDUP(matname);
     mat->mode = mode;
 
     if ( mat->version == 0x0200 ) {
@@ -619,7 +619,7 @@ Mat_VarCreate(const char *name,enum matio_classes class_type,
     matvar->isGlobal    = opt & MAT_F_GLOBAL;
     matvar->isLogical   = opt & MAT_F_LOGICAL;
     if ( name )
-        matvar->name = strdup_printf("%s",name);
+        matvar->name = STRDUP(name);
     matvar->rank = rank;
     matvar->dims = NEW_ARRAY(size_t,matvar->rank);
     for ( i = 0; i < matvar->rank; i++ ) {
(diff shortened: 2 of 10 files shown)


papadop:
Cannot understand the travis failure. For me the test is skipped (as is the uncompressed case), so I do not understand why the test is not skipped there... Used valgrind to check for memory errors and found nothing.... I can understand that the patch could have introduced a uninitialized memory problem (forgot to say that some calloc where replaced by mallocs and proper memsets to initialize the memory where needed). 

Is there a way to access the travis build ???

tbeu:
One test failed: mat5_compressed_write.at:573
What is stdout and stderr for command `test_mat -v 5 -z write_2d_logical`?

papadop:
sipadan-> ./test_mat -v 5 -z write_2d_logical
sipadan-> echo $status
0

papadop:
So no stdout or stderr.

sipadan-> ls -la test_write_2d_logical.mat 
-rw-r--r-- 1 papadop athena 405 Mar 29 16:16 test_write_2d_logical.mat
sipadan-> more test_write_2d_logical.mat
MATLAB 5.0 MAT-file, Platform: x86_64-unknown-linux-gnu, Created by: libmatio v1
.5.6 on Tue Mar 29 16:16:30 2016
[...]

papadop:
The test failure is very very probably due to the missing \n. The other pull request should correct that... I did not made the change in this branch to avoid creating conflicts. The mystery is more why this test is not skipped...

papadop:
Remember that this PR is more for you to see where I'm going and provide you (and me) a chance of providing comments early. This is WIP. Note that mat4.c is now mostly totally converted to try/catch. So you can see from this branch what the code will look like in the end if this idea is pursued. Some breakage is to be expected for the time being. Will remove them for the actual (real) PR.

agramfort:
FYI https://github.com/openmeeg/matio-openmeeg/pull/4/files
here is working travis setup using cmake for building and testing on mac os + linux

tbeu:
Matio is a pure C library and not designed as C++ library. See the replicated ReadData function that only differ on the data type and could be very conveniently written using C++ templates. E.g. the [msgpack-c](https://github.com/msgpack/msgpack-c) project has a true C and a true C++ implementation that not just differs on mallo/free vs. new/delete or file I/O. Therefore I propose to keep Matio as C-only library. And I know there are plenty of malloc's with unchecked return values within Matio that need to be tackled.

agramfort:
ok fair enough but is there a proper/clean way for us to catch malloc
issues from a c++ project?

tbeu:
Usually malloc errors should be caught within the library. Do you call Mat_LogInit in your application such that allocation errors are reported/dealt with?

tbeu:
Closing...

PR #36, opened 2016-05-11
API change

papadop:
If we want people to be able to write their own log functions, we have to expose the meaning of the log levels. This patch does that . The names are changed to protect them from any conflicts with other libraries. All (non matlab) tests pass.


tbeu:
ci.appveyor.com failed but could be due some changed download links :8ball:

PR #42, opened 2016-08-24
Eliminate use of mktemp -- use mkstemp instead

gsjaardema:
The use of mktemp() results in an unsuppressable warning message
    on certain compilers, including gcc.  This causes problems for
    projects which have rules regarding warning-free compiles.

    Use mkstemp instead. The problem with mkstemp is that it opens the
    file instead of just returning a unique file name.  However, as a
    side effect it does change the template to a unique file name.
    This can be used similarly to the return value from mktemp.


tbeu:
Thanks. Fails on Win though: https://ci.appveyor.com/project/tbeu/matio/build/1.5.8+build.14#L303 Need to check.

PR #43, opened 2016-08-24
Eliminate use of mktemp -- use mkstemp instead

gsjaardema:
The use of mktemp() results in an unsuppressable warning message
on certain compilers, including gcc.  This causes problems for
projects which have rules regarding warning-free compiles.

Use mkstemp instead. The problem with mkstemp is that it opens the
file instead of just returning a unique file name.  However, as a
side effect it does change the template to a unique file name.
This can be used similarly to the return value from mktemp.


tbeu:
Why not stay with mktemp on MSVC as e.g. here: https://github.com/tkelman/graphviz/commit/acdbce81a005e88fd6fffb998c97d27e6e1d303f ?

gsjaardema:
That is a valid alternative. To my eye, it dirties up the actual source code more with all the ifdefs in the code versus an ifdef up at the top in the "preamble" code.

Perhaps a better fix than either would be to encapsulate the mkstemp/mktemp changes in a static function.  The function would return the filename that should be used in the Mat_CreateVer call.  This would hide all ifdefs at the top and eliminate confusion as to why mkstemp() was being used just for the side-effect of changing the filename template.

src/mat.c
```diff
@@ -883,7 +885,8 @@ Mat_VarDelete(mat_t *mat, const char *name)
                 break;
         }
 
-        tmp = Mat_CreateVer(tmp_name,mat->header,mat_file_ver);
+        close(fd);
+        tmp = Mat_CreateVer(temp,mat->header,mat_file_ver);
```
tbeu: This introduces a race condition: After closing the file descriptor `fd` of the temporary file `temp` might no longer be a unique temporary filename. Thus a proper fix also requires replacing all `fwrite` calls by ``write` (which is some effort).

src/mat.c
```diff
@@ -52,31 +52,36 @@
...
+  static char template[7] = "XXXXXX";
+  static char temp[7];
+  strcpy(temp, template);
+#if defined(_WIN64) || defined(_WIN32)
+  return _mktemp(temp);
+#else
+  int fd = mkstemp(temp);
+  if (fd > 0) {
+    close(fd);
+    return temp;
```
tbeu: This introduces a race condition: After closing the file descriptor `fd` of the temporary file `temp` might no longer be a unique temporary filename. Thus a proper fix also requires replacing all `fwrite` calls by `write` (which is some effort and changes the existing API functions like `Mat_CreateVer`).

gsjaardema: True, but I guess I was going for good enough vs perfect.  The current pull request eliminates the compiler warnings from mktemp and is at least as good with regards to security and race conditions.

tbeu: Note that this is an old issue: https://sourceforge.net/p/matio/bugs/10/
I already considered to fix it but want to do it the perfect way.

gsjaardema: Not sure that the fwrite vs write is the perfect fix either since I don't think it will work with the hdf5-based file formats will it?

tbeu: Just saw http://stackoverflow.com/q/1941464. ANSI C compatibility also is important (as I was told at https://trac.modelica.org/Modelica/ticket/1258).

tbeu:
Without the dependency on the hdf5 library I figured out a secure temp file creation in Mat_VarDelete using tmpfile (and thus avoiding the file name argument). However, as long as the hdf5 library does not support any secure temp file creation and management, I will not touch it. (HDF5 issue number is HDFFV-8703 from 2014.)

So, for now, I close this with wontfix.

PR #50, opened 2016-12-05
update README installation instructions

cs01:
Add steps to run `autoconf.sh` and update the `test/datasets` submodule

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,10 @@ For Windows, the pre-compiled binaries can be used which also include a DLL of z
 #### 2.2.1 Quick Build Guide
 The primary method for building the software is using `configure` followed by `make`. After building, the testsuite can be executed to test the software using `make check`. The software can be installed using `make install`. For example,
 ```sh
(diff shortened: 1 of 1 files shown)


tbeu:
Thanks for the PR. I actually added it by 2f0a6ebc2aff67c52623d83ac1ddb947fb4021f4 before I removed it again by 6e61ad908746351c818da65fdea532e495031187. Maybe I need to differentiate between the README.md on GitHub and the README as part of the source distribution.

cs01:
Agreed. I burned more time than I'd like to admit because I didn't know I was supposed to run the autoconf script or update submodules.

Also, I don't know the history of matio's hosting on sourceforge vs. github, but I found it confusing that there are two locations git repositories, two issue tracking systems, etc. Is it be possible to only have a single home for matio on the internet?

tbeu:
For historical reasons matio is hosted at sf.net. But as GitHub offers better CI possibilities I have matio mirrored at this GitHub repo and keep both repos in sync.

PR #63, opened 2017-06-02
Add 'autoconf' to 'Quick Build Guide'

NiklasEi:
For someone strictly following the 'Quick Build Guide' running './configure' will produce 'bash: ./configure: No such file or directory'. To spare others the same problem 'autoconf' was added to the guide which will produce the missing configure script with the given 'configure.ac'.

I did not change anything on line 42 but GitHub automatically changed the encoding of the file from ISO-8859-1 to UTF-8, which I guess is leading GitHub to mark the line as changed.

### DIFF
--- a/README
+++ b/README
@@ -39,7 +39,7 @@ Table of Contents
           * Nils Jannasch (https://github.com/NJannasch)
           * OpenMEEG ((http://openmeeg.github.io/)
           * Scilab (http://www.scilab.org/)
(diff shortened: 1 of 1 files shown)


tbeu:
Note, there is a difference between building the lib from the repo or building from the distribution. README is meant for the distribution, where configure is available.

tbeu:
I fixed the encoding issue in both READMEs though. Thanks for reporting to it.

NiklasEi:
Sure, no problem.

PR #72, opened 2017-11-28
Convert to pragma once.

papadop:
Simple change to convert .h files to #pragma once.
All modern C compilers I know support this. 
But there is the Portland compiler that matio seems to support (I'm not sure how many builds are made with this compiler).
https://en.wikipedia.org/wiki/Pragma_once
Anyway, I'm putting this here for reference.
All test pass.

### DIFF
--- a/getopt/getopt.h
+++ b/getopt/getopt.h
@@ -37,8 +37,7 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-#ifndef _GETOPT_H_
-#define _GETOPT_H_
+#pragma once
 
 #if !defined(_WIN32) && !defined(_WIN64)
 #   include <sys/cdefs.h>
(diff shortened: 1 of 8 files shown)


tbeu:
There is no real need/requirement to change this and I'd prefer to leave it as it is, i.e., not to merge. OK for you? See also https://stackoverflow.com/questions/1143936/pragma-once-vs-include-guards.

papadop:
Yes. This is OK. A in the stackoverflow comment, I tend to prefer pragma once over standard include guard as they are less error prone and less verbose. It was easy to do and I did it on the side of the cmake port (by the way compilation works, I need to do some work to on tests, I'll publish a PR - just for comments -- soon). I discovered only after the fact that PGI compilers do not support it (and they are the only ones, which is a shame.....).

tbeu:
Agreement to close without merge.

PR #73, opened 2017-11-29
Cmake

papadop:
This is a preview of the cmake integration.... I verified that it both compiles with autoconf and cmake.
tests are not yet fully integrated within the cmake port. There are missing tests and 153 of te 2022 cmake tests are failing (probably because the tests are wrongly written).

This is only for getting comments for now....


papadop:
Oups I obviously need to rebase on master....

tbeu:
You still can do it in your branch and `push --force` which automatically updates this PR.

Issue #51, opened 2017-02-08
Fortran interface for matio


onyukim:
Hello,

I have found that Fortran interface for matio cannot be compiled.
I would be grateful if you could fix that.

tbeu:
I am afraid the Fortran interface is not well maintained. I would be very glad if you could figure out what the problem is.

onyukim:
Thank you for the response.

- Configurations for `src/fortran/Makefile.am` and `src/fortran/matio_t.inc.in` are missing in `configure.ac`
- Configurations for a Fortran compiler is also missing in that file.

I added some lines as shown below. However, I don't know whether it is correct because of my lack of knowledge about `autoconf` and `automake`.
```diff
diff --git a/configure.ac b/configure.ac
index 5631856..d353c6b 100644
--- a/configure.ac
+++ b/configure.ac
@@ -56,6 +56,7 @@ dnl  Build programs, C compiler, F77 compiler, make, install, etc.
 dnl
 AC_PROG_CC([pgcc icc gcc cc])

+AC_PROG_FC(ifort)
 AC_ARG_ENABLE(coverage,
 [  --enable-coverage       Enable coverage testing],
 [CFLAGS="$CFLAGS -fprofile-arcs -ftest-coverage"])
@@ -356,6 +357,8 @@ AC_SUBST([AUTOTEST], ['$(AUTOM4TE) --language=Autotest'])
 AC_CONFIG_FILES([Makefile \
                  matio.pc \
                  src/Makefile \
+                 src/fortran/Makefile \
+                 src/fortran/matio_t.inc
                  getopt/Makefile \
                  tools/Makefile \
                  test/Makefile \
```

After `./autogen.sh` and `./configure`, the derided files was generated, but `make` in `src/fortran`  failed to compile the interface.
I attached the message given by a compiler (I use Inter C and Fortran compilers).

[message_from_compiler.txt](https://github.com/tbeu/matio/files/763098/message_from_compiler.txt)

tbeu:
fd2ca03d16ff7e2ed8fa7c56da89fec922eefd7a removed the Fortran support from the configure and build options. I do not know why it was removed and will see if I get it running again.

tbeu:
There is now a branch [fortran](https://github.com/tbeu/matio/tree/fortran) with reverted fd2ca03d16ff7e2ed8fa7c56da89fec922eefd7a. However it does not build. Also, the original developer of matio did not yet reply on my mail asking for support.

[...]

PR #74, opened 2017-11-29
New cmake branch

papadop:
This is only WIP for now. All comments accepted !
The cmake build system is added. I verified it build both in old fashion and with cmake.
Tests with cmake are currently both incomplete (2018 tests instead of 3152) and sometimes wrong (180 tests out of 2018 are failing). I'm working on it.


papadop:
appveyor build fails, but I do not think I'm responsible....

emmenlau:
I'm very cuirous about this, thanks!

tbeu:
I restarted the appveyor build.

papadop:
Some options are also missing.

tbeu:
Why did you introduce matio_noreturn.h?

papadop:
To factor this piece of code between the matio_pubconf.h.in and matio_pubconf_cmake.in. In this way the two files only contain defines (and includres). It is also simpler to check the correspondance between the defined variables (which by the way I need to check if there were updates in this respact since I did the original cmake port).

tbeu:
I still prefer to one single pubconfig file and not distributed config files.

emmenlau:
On MSVC I get an error that the folder `getopt` does not contain a CMakeLists.txt file.

emmenlau:
Ok, I found a simple `CMakeLists.txt` for getopt. Now I'm still stuck at:
```
snprintf.c.obj : error LNK2019: unresolved external symbol va_copy referenced in function rpl_vasprintf
```
Are there extra libraries that need to be linked on Windows? The cmake-generated `matioConfig.h` defines `HAVE_VA_COPY` and undefines `HAVE___VA_COPY`, so I think that its correct.

emmenlau:
Please consider PR https://github.com/tbeu/matio/pull/75 that addresses several of the above problems.

papadop:
I just pushed the one from OpenMEEG-matio... It should work out of the box as we regularly build this version of matio on windows...

cmake/Packaging.cmake
```diff
@@ -0,0 +1,116 @@
...
+            set(PACKAGE_NAME ${PACKAGE_NAME}-shared)
+        else()
+            set(PACKAGE_NAME ${PACKAGE_NAME}-static)
+        endif()
+
+        set(CPACK_PACKAGE_FILE_NAME ${PACKAGE_NAME})
+
+        if (WIN32)
+            # There is a bug in NSIS that does not handle full unix paths properly. Make
+            # sure there is at least one set of four (4) backlasshes.
```
tbeu: Typo: backslashes

cmake/Packaging.cmake
```diff
@@ -0,0 +1,116 @@
...
+
+if (NOT PACKAGE_NAME)
+    if (ENABLE_PACKAGING OR BUILD_RPM)
+        include(InstallRequiredSystemLibraries)
+
+        configure_file(${MATIO_SOURCE_DIR}/README    README      COPYONLY)
+        configure_file(${MATIO_SOURCE_DIR}/README.md README.md   COPYONLY)
+        configure_file(${MATIO_SOURCE_DIR}/COPYING   COPYING.txt COPYONLY)
+
+        set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "matio library for matlab IOs")
```
tbeu: Preferred: MATLAB MAT file I/O

cmake/Packaging.cmake
```diff
@@ -0,0 +1,116 @@
...
+        set(CPACK_PACKAGE_VENDOR "Christopher Hulbert")
+        set(CPACK_PACKAGE_DESCRIPTION_FILE "${MATIO_BINARY_DIR}/README.md")
+        set(CPACK_RESOURCE_FILE_LICENSE "${MATIO_BINARY_DIR}/COPYING.txt")
+        set(CPACK_PACKAGE_INSTALL_DIRECTORY "matio")
+        set(CPACK_PACKAGE_CONTACT "cch@isl-inc.com")
```
tbeu: Better use the sf mail address

cmake/Packaging.cmake
```diff
@@ -0,0 +1,116 @@
...
+
+        if (UNIX AND BUILD_RPM) # linux
+            set(CPACK_GENERATOR "${CPACK_GENERATOR};RPM")
+            if (CMAKE_VERSION VERSION_LESS "2.8")
+                include(UseRPMTools)
+                if (RPMTools_FOUND)
+                    RPMTools_ADD_RPM_TARGETS(${PROJECT_NAME} "${PROJECT_SOURCE_DIR}/packaging/${PROJECT_NAME}.spec.in")
+                endif()
+            else()
+                set(CPACK_RPM_PACKAGE_LICENSE "LGPL")
```
tbeu: It is the BSD 2-Clause License now.

src/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+    set(SRC_FILES ${SRC_FILES} inflate.c)
+endif()
+
+if (WIN32)
+    set(SRC_FILES ${SRC_FILES} matio.rc)
+endif()
+
+set(FORTRAN_INTERFACE_FILE)
+if (ENABLE_FORTRAN)
+    include(FortranCInterface)
```
tbeu: Fortran interface is unmaintained, see #51.

src/matio.h
```diff
@@ -230,95 +230,95 @@ typedef struct mat_sparse_t {
 /** @endcond */
 
 /* Library function */
-EXTERN void Mat_GetLibraryVersion(int *major,int *minor,int *release);
+EXTERN MATIO_EXPORT void Mat_GetLibraryVersion(int *major,int *minor,int *release);
```
tbeu: Please try to be minimal invasive on existing files. EXTERN alone should be enough as it can also handle the export definition.

test/CMakeLists.txt
```diff
@@ -0,0 +1,151 @@
+add_executable(test_mat test_mat.c)
+target_link_libraries(test_mat matio ${GETOPT_LIB})
+
+add_executable(test_snprintf test_snprintf.c ${MATIO_SOURCE_DIR}/src/snprintf.c)
+target_link_libraries(test_snprintf ${GETOPT_LIB} ${LIBM})
+
+option(MATLAB_TESTING "Enable matlab read tests (requires a function matlab)" ON)
```
tbeu: Whenever possible, type MATLAB in capitals.

test/testsuite.at
```diff
@@ -95,11 +95,11 @@ fi
 ]
 m4_divert_pop([PREPARE_TESTS])
 
-AT_BANNER(Read [MAT version 4 (Little Endian)])
+AT_BANNER([Read MAT version 4 (Little Endian)])
 m4_include([mat4_read_le.at])
 m4_include([mat4_readslab_le.at])
 
-AT_BANNER(Read [MAT version 4 (Big Endian)])
+AT_BANNER([Read MAT version 4 (Big Endian)])
```
tbeu: These changes are independent of the Cmake stuff and should be an extra PR - if you like.

tbeu: Why is it necessary to move all at files? I'd prefer to keep it as is.

papadop: At the beginning, I was considering creating a cmake testsuite mimicking the autotools one. That would have made a lot of supplementary files and the directory would have been cluttered. So I put all the testsuite files (besides the main one) in their own directory. Then, I knew it would be difficult to maintain two testsuites in parallel (and I saw your comment of the matio-pubconf.h file), so I went the route of trying to parse the current testsuite (which is somewhat fragile as I'm doing pattern matching, so any change in the testsuite might create problems but seems to work). 

Yet, I find the organization of the testsuite in a subdir cleaner...

At this point, I'm first trying to have everything working and in place with at least one travis build working (have not found appveyor configuration). If you insist on keeping the files in their original place, I will do it, but I think the reorganization is somewhat better...

papadop: For the other PR, I afraid of conflicts. But if you prefer, I  surely can do that.

test/CMakeLists.txt
papadop: OK done. Will commit later.

test/testsuite.at
tbeu: The Appveyor config is only in the cloud. It basically calls msbuild on the VS2010 solution.

This already looks like a monster PR. Keeping some independent changes in its own PR might help for clarity.

tbeu: > If you insist on keeping the files in their original place, I will do it, but I think the reorganization is somewhat better...

Not sure what to prefer. But as maintainer I was rather conservative here and tried to avoid any structural changes to the original code.

Does https://cmake.org/Wiki/CMakeEmulateMakeCheck give some clue?

src/matio.h
papadop: EXTERN will also export all the other routines defined in mat4.c mat5.c and mat7.3.c. So it's not appropriate I think. The routines marked with MATIO_EXPORT are those mentioned in matio.def... Maybe the best way to fuse both autoconf and cmake approaches would be to use the cmake way of marking exported routines with this MATIO_EXPORT macro. This has the nice advantage that visibility is directly controlled in one place (and explicitly visible to the user). I do not know how to change the VisualStudio project files to do that, but I can produce the proper matio_export.h from autoconf if I can convince you this is a better route to follow... ;-)

After seeing your comment, I saw a way to add the matio.def file in cmake. But honestly, nowadays marking the routines in the source file seems to be somewhat preferred (all the more that it adds visibility control for gcc which does not have the .def equivalent).

src/CMakeLists.txt
papadop: OK. I must say that I have not tested it much recently. If the cmake version can help here, why not. If you prefer to totally forget the fortran interface, I'll happily will remove that code...

cmake/Packaging.cmake
papadop: Done, will appear with next push.

cmake/Packaging.cmake
papadop: Done.

cmake/Packaging.cmake
papadop: Done, I have put the two sf addresses I found in README.md. 
In the long run a mailist might be better....

src/matio.h
tbeu: Hm, in 001704f2425c0e86c3139c6c5651f0f2a9286667 I added the matio.sym file to fix the visibility control for gcc.

cmake/Packaging.cmake
papadop: Done

tbeu:
Is this still being worked on? If not, I propose to close it w/o merged?

tbeu:
> Is this still being worked on? If not, I propose to close it w/o merge?

Doing so.

PR #85, opened 2018-03-10
Move tests into their own directory to cleanup the test directory.

papadop:
I know that you were not so keen to this change in the cmake branch, but I think it make the test directory cleaner and more readable, so I attempted to isolate this change and propose it separately from the cmake branch. In addition, test/Makefile.am is slightly improved by adding some variables, which make it -- in a minor fashion as I do not expect dirname changes so often -- easier to maintain...

Fully tested on linux fedora-27 x86-64 with no regressions (no matlab tests, but I do not see how those would specially fail when all others are succeeding).


tbeu:
I guess you forgot to adapt testsuite.at.

papadop:
Strange, it worked on my computer. There is in theory no testsuite.at change as we just specify a test include directory in the Makefile.am. I'll investigate....

papadop:
Indeed I forgot to update the -I in the autotest command line. The last commit should correct it. I dp not know why it worked when I tested it. Maybe I used a wrong repository (I have many) or forgot to update with autoreconf. Let's see if tarvis passes now. Note that this also shows one advantage of cmake build. When it is finished the testsuite will also work for windows builds.... I just realized yesterday that the testsuite was unix/macos only....

PR #86, opened 2018-03-13
Sizeof cleanup

papadop:
This is a cleanup of the code which removes the need to check for type sizes in the configure step.
It also factorizes some code and probably more readable.

I have not yet suppressed for configure.ac:
- SIZZEOF_VOID_P which is used in the testsuite, but we probably can generate a shell script which would avoid the grepping in matioConfig.h (so faster and cleaner).
- SIZEOF_SIZE_T because it is used in the fortran interface (yes I know it does not work, but I do not want to worsen the situation with this patch).

My short term goal is to simplify as much as possible matioConfig.h.in so that only defines used for matio remain. This also means that I will separate the defines only needed for snprintf.c in another file (probably my next task).

Tested on a linux 64 bits computer.

Let's see if the builders are happy...


papadop:
Another nice feature of this patch is that some previously unchecked configuration will now generate a runtime error (I must admit that there is close to zero chance that these cases are met: sizeof pointer different from 4 or 8, or size of  a type not covered by char, short, int, long and long long or their unsigned counterparts).

papadop:
OK some small issues corrected (windows visual C  not knowing inline and a typo which was generating a warning which I had not seen).

src/mat.c
```diff
@@ -1691,15 +1691,17 @@ Mat_VarGetSize(matvar_t *matvar)
...
+        /* 112 bytes cell/struct overhead for 64-bit system */
+        overhead = 112;
+        ptr = 8;
+    } else if (sizeof(void*)==4) {
+        /* 60 bytes cell/struct overhead for 32-bit system */
+        overhead = 60;
+        ptr = 4;
+    } else {
+        Matio_Error("Pointers of size %d are not supported by matio.",sizeof(void*)*CHAR_BIT);
+    }
```
tbeu: This changes a compiletime check by a run-time check. I'd rather have it as it is as it already uses compiler defines used by MSVC or GCC/clang.

src/endian.c
```diff
@@ -214,35 +214,38 @@ Mat_floatSwap( float *a )
...
+            swap( tmp.a[3], tmp.a[12] );
+            swap( tmp.a[4], tmp.a[11] );
+            swap( tmp.a[5], tmp.a[10] );
+            swap( tmp.a[6], tmp.a[9] );
+            swap( tmp.a[7], tmp.a[8] );
+            break;
+        default:
+            Mat_Error("The double type of this architecture (%d) is not handled",sizeof(double));
+    }
+
```
tbeu: This changes compile-time checks by runtime comparisons. I see no reason to change it.

papadop: Any decent compiler will optimize it as sizeof(double) is a constant and the tests are then tautological. Even the default should be eliminated.  Yet there is a "security" which was not existing before just in case ...
strings endian.o | grep "The double type" returns nothing." returns nothing which indeed shows that the call Mat_error has been eliminated. The only improvement I can see would be to catch this situation at configuration time. This was not done before and could be added.... So the patch introduces no regression and no execution time penalty as the switch is optimized away.

papadop:
There is a problem with matioConfig.h.in ... Probably related to some .gitignore issue, because git tells me that the file is not tracked but was obtained when I checked out the repository.... I will correct that. By the way I have several files generated by autoconf/automake which appear in the status. Would it make sense to add them to .gitignore ? My autoconf/automake mastership is not what it was a few years ago, so I may be wrong....

bali-> git status
On branch SIZEOF_cleanup
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   config/config.guess
	modified:   config/config.sub
	deleted:    config/depcomp
	modified:   config/install-sh
	modified:   config/missing

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	config/compile
	src/matioConfig.h.in

no changes added to commit (use "git add" and/or "git commit -a")

tbeu:
That is what I see on openSUSE Leap 42.3:
```sh
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
Untracked files:
  (use "git add <file>..." to include in what will be committed)

        matio.pc
        src/matio_pubconf.h
        src/stamp-h2
        test/atconfig

nothing added to commit but untracked files present (use "git add" to track)
```

tbeu:
Closing now, since there is no more progress to expect.

PR #93, opened 2018-10-02
Write backwards compatible Mat7.3 files

StefanBruens:
When Matio uses HDF5 1.10.x, it may create files using features only
supported with newer versions.
Limiting the version range for compatibility was added in HDF5 1.10.2.
For details, see
https://www.hdfgroup.org/2018/04/why-should-i-care-about-the-hdf5-1-10-2-release/

### DIFF
--- a/src/mat73.c
+++ b/src/mat73.c
@@ -2200,11 +2200,15 @@ Mat_Create73(const char *matname,const char *hdr_str)
     mat_t *mat = NULL;
     size_t err;
     time_t t;
-    hid_t plist_id,fid;
+    hid_t plist_id,fid,plist_ap;
 
     plist_id = H5Pcreate(H5P_FILE_CREATE);
     H5Pset_userblock(plist_id,512);
-    fid = H5Fcreate(matname,H5F_ACC_TRUNC,plist_id,H5P_DEFAULT);
+    plist_ap = H5Pcreate(H5P_FILE_ACCESS);
+#if H5_VERSION_GE(1,10,2)
+    H5Pset_libver_bounds(plist_ap,H5F_LIBVER_EARLIEST,H5F_LIBVER_V18);
+#endif
+    fid = H5Fcreate(matname,H5F_ACC_TRUNC,plist_id,plist_ap);
     H5Fclose(fid);
     H5Pclose(plist_id);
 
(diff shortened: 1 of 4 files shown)


tbeu:
Thanks for the PR. I wonder why the new lines https://coveralls.io/builds/19311752/source?filename=src/mat73.c#L2209 and https://coveralls.io/builds/19311752/source?filename=src/mat.c#L415 are not covered. Does H5F_LIBVER_V18 need to be set explicitly?

StefanBruens:
> Thanks for the PR. I wonder why the new lines https://coveralls.io/builds/19311752/source?filename=src/mat73.c#L2209 and https://coveralls.io/builds/19311752/source?filename=src/mat.c#L415 are not covered. Does H5F_LIBVER_V18 need to be set explicitly?

Unfortunately, the preprocessor check was wrong, H5F_LIBVER_V18 is an enum value and thus undefined from the preprocessors point of view.

Changed the check to H5_VERSION_GE(1,10,2), which has the intended effect.

src/mat73.c
```diff
@@ -2200,11 +2200,15 @@ Mat_Create73(const char *matname,const char *hdr_str)
     mat_t *mat = NULL;
     size_t err;
     time_t t;
-    hid_t plist_id,fid;
+    hid_t plist_id,fid,plist_ap;
 
     plist_id = H5Pcreate(H5P_FILE_CREATE);
     H5Pset_userblock(plist_id,512);
-    fid = H5Fcreate(matname,H5F_ACC_TRUNC,plist_id,H5P_DEFAULT);
+    plist_ap = H5Pcreate(H5P_FILE_ACCESS);
```
tbeu: plist_ap is not closed in case of the two early returns in line 2217 and 2224.

StefanBruens: Whats preferable - closing in the error path, or closing always and recreating the FAPL prior to H5Fopen?

tbeu: Hm, I have no strong opinion here, but closing in the error path sounds ok.

tbeu:
Resolved by cc80e26cee1665ff8f0afae81268e75257f9f39d. Thanks.

StefanBruens:
Sorry for not fixing it, completely forgot to...

PR #97, opened 2018-11-05
Avoid using mktemp (considered harmful)

timangus:
At link time, GCC emits the following: `warning: the use of 'mktemp' is dangerous, better use 'mkstemp'`. This patch uses `mkstemp` intead, as recommended.


tbeu:
This is a duplicate of of #43.

PR #115, opened 2019-04-02
test against hdf5 shipped in brew

massich:


tbeu:
@massich Do you plan to fix the travis config?

massich:
It needed fixing in brew, and as far as I can tell it stall. I'm not sure if it's still broken.

tbeu:
Can you rebase and force-push to trigger travis again?

tbeu:
Still the same config problem that it does not built with hdf5 support.

tbeu:
@massich How to proceed with this PR?

tbeu:
Works when xcode 9.4 is used (instaed of 8.3).

PR #125, opened 2019-09-09
Fix so prototype matches implementation

gsjaardema:
The argument type of one of the arguments in the declaration of `ReadSparse` was not changed when the corresponding argument type was changed at the implementation site.


tbeu:
Already fixed by 32ca06b826d04d89a82be40ffce1ad4236d61f5f.

PR #132, opened 2019-12-05
fix mktemp linker warnings when using GNU linker

stephen-sorley:
Use mkdtemp() to create a temporary directory with a unique name,
then add a constant filename to the path after that. This allows
us to construct a unique path without any race conditions, and
without having to use mktemp().

Hopefully doing it this way avoids the issues you were having
with getting mkstemp() to work in #43.

### DIFF
--- a/src/mat.c
+++ b/src/mat.c
@@ -61,22 +61,35 @@
  *===================================================================
  */
 
+#define MAT_MKTEMP_DIR "/tmp/"
+#define MAT_MKTEMP_TPL "XXXXXX"
+#define MAT_MKTEMP_FILE "/temp.mat"
+
+static const size_t MAT_MKTEMP_BUF_SIZE =
+  sizeof(MAT_MKTEMP_DIR)+sizeof(MAT_MKTEMP_TPL)+sizeof(MAT_MKTEMP_FILE)-2;
+
 static char *
-Mat_mktemp(char *temp)
+Mat_mktemp(char *path_buf, char *dir_buf)
 {
     char *ret = NULL;
 
+    *path_buf = '\0';
+    *dir_buf  = '\0';
+
 #if (defined(_WIN64) || defined(_WIN32)) && !defined(__CYGWIN__)
-    if(_mktemp(temp) != NULL) {
-        ret = malloc(strlen(temp) + 1);
-        strcpy(ret, temp);
-    }
+    strcpy(path_buf, MAT_MKTEMP_TPL);
+    if ( NULL != _mktemp(path_buf) )
+        ret = path_buf;
 #else
-    const char *name = "/temp.mat";
-    if(mkdtemp(temp) != NULL) {
-        ret = malloc(strlen(temp) + strlen(name) + 1);
-        strcpy(ret, temp);
-        strcat(ret, name);
+    /* On Linux, using mktemp() causes annoying linker errors that can't be
+       suppressed. So, create a temporary directory with mkdtemp() instead,
+       and then just always use the same hardcoded filename inside that temp dir.
+     */
+    strcpy(dir_buf, MAT_MKTEMP_DIR MAT_MKTEMP_TPL);
+    if ( NULL != mkdtemp(dir_buf) ) {
+        strcpy(path_buf, dir_buf);
+        strcat(path_buf, MAT_MKTEMP_FILE);
+        ret = path_buf;
     }
 #endif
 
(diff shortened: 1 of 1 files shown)


src/mat.c
```diff
@@ -1171,7 +1192,7 @@ Mat_VarDelete(mat_t *mat, const char *name)
     if ( NULL == mat || NULL == name )
         return err;
 
-    if ( (tmp_name = mktemp(temp)) != NULL ) {
+    if ( (tmp_name = Mat_mktemp(temp)) != NULL ) {
```
tbeu: Does the template needs to be adpated from "XXXXXX" to "/tmp/tmpdir.XXXXXX"?

src/mat.c
```diff
@@ -1276,6 +1297,7 @@ Mat_VarDelete(mat_t *mat, const char *name)
                 Mat_Critical("Cannot remove file \"%s\".",tmp_name);
             }
         }
+        free(tmp_name);
```
tbeu: In one of my use-cases I "misuse" Mat_Critical to never return to the caller (using setjmp/longjmp). In that case, the freeing is too late here.

src/mat.c
```diff
@@ -62,6 +61,28 @@
...
 
+static char *
+Mat_mktemp(char *temp)
+{
+    char *ret = NULL;
+
+#if (defined(_WIN64) || defined(_WIN32)) && !defined(__CYGWIN__)
+    if(_mktemp(temp) != NULL) {
+        ret = malloc(strlen(temp) + 1);
+        strcpy(ret, temp);
```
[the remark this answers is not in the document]
stephen-sorley: Fixed by having the caller pass in a stack-allocated buffer instead.

src/mat.c
```diff
@@ -62,6 +61,28 @@
...
+    }
+#else
+    const char *name = "/temp.mat";
+    if(mkdtemp(temp) != NULL) {
+        ret = malloc(strlen(temp) + strlen(name) + 1);
```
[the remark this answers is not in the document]
stephen-sorley: Fixed by having the caller pass in a stack-allocated buffer instead.

src/mat.c
stephen-sorley: I don't think the old implementation with mktemp() was putting the temporary files in /tmp - I think it was just adding them to the current working directory (unless I misread the code).

However, I went ahead and made this change, because it seems like a good idea to put temporary files in /tmp anyway.

I also added code to delete the temporary directory (in addition to deleting the temporary file), in case that was why you suggested putting it in /tmp.

src/mat.c
stephen-sorley: Changed to have the caller stack-allocate buffers and pass them in, to avoid having to bother with free().

tbeu:
@svillemot The changes look very good to me now. Before merging I'd like to get your feedback. Do you see any issues with mkdtemp on Debian platforms? Should we enable mkdtemp only for POSIX systems? Thanks a lot.

svillemot:
No problem for using `mkdtemp()` on Debian. But you may indeed have to disable it for non-POSIX systems (for example, I don’t think it is available under MinGW).

stephen-sorley:
@tbeu Any further changes that you need me to make to this?

What platforms do you support besides Windows/Linux/macOS? Apart from Windows, looks like AIX is the main one that's missing mkdtemp:
http://lists.llvm.org/pipermail/llvm-dev/2010-August/034081.html

Solaris provides it, however:
https://docs.oracle.com/cd/E36784_01/html/E36874/mkdtemp-3c.html

tbeu:
> @tbeu Any further changes that you need me to make to this?

No, it all looks fine. I only was reluctant to merge since @svillemot mentioned MinGW. I need to give it one more try after I observed no issues when I tested it locally last time.

tbeu:
> I only was reluctant to merge since @svillemot mentioned MinGW. I need to give it one more try after I observed no issues when I tested it locally last time.

I rebased (and force-pushed) your bracnh to confirm that it builds with MinGW.

> What platforms do you support besides Windows/Linux/macOS? Apart from Windows, looks like AIX is the main one that's missing mkdtemp: http://lists.llvm.org/pipermail/llvm-dev/2010-August/034081.html

I have no chance to build on AIX. I believe we should exclude it from the two readme files, where AIX is mentioned as supported platform. OK for you?

stephen-sorley:
Yeah, excluding AIX sounds good to me. I don't have any way to test on it, either.

tbeu:
Thanks for the pull request.

PR #149, opened 2020-08-26
Update mat5.c to support uncompressed MAT file generated by SWAN 

dpzimmer:
Slight modification to Mat_VarReadNextInfo5() to allow MAT_T_INT32 in addition to MAT_T_UINT32 to indicate processing of array_flags.

This formatting was encountered processing older .mat files created by older versions of the swan model (https://www.tudelft.nl/en/ceg/about-faculty/departments/hydraulic-engineering/sections/environmental-fluid-mechanics/research/swan/).

I have no idea if this violates the file format specification, but Matlab and libmat.lib both read such files without complaint.

### DIFF
--- a/src/mat5.c
+++ b/src/mat5.c
@@ -5282,7 +5282,7 @@ Mat_VarReadNextInfo5( mat_t *mat )
             }
 
             /* Array flags */
-            if ( buf[0] == MAT_T_UINT32 ) {
+            if ( buf[0] == MAT_T_UINT32 || buf[0] == MAT_T_INT32 ) { /* Also allow INT32 for SWAN */
                array_flags = buf[2];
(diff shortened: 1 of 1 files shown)


tbeu:
Thanks for your contribution.

That particular check occurs six times in mat5.c and I guess, it needs to be fixed everywhere consistently.

Do you have an example file that you could share. Thanks.

dpzimmer:
> Thanks for your contribution.
> 
> That particular check occurs six times in mat5.c and I guess, it needs to be fixed everywhere consistently.
> 
> Do you have an example file that you could share. Thanks.


[swan.zip](https://github.com/tbeu/matio/files/5132581/swan.zip)

tbeu:
Thanks for the MAT file.

I also checked the sources of SWAN and it is exactly that one relevant line that needs to be adapted.

PR #151, opened 2020-09-05
CMake improvements

MaartenBent:
This fixes using the `hdf5-config.cmake` provided by the HDF5 package. This didn't work correctly because it checked for an exact version match.

I now automatically select static HDF5 (and zlib/szip) libraries when building static matio. ~Another idea is adding an option `MATIO_HDF5_LIBRARY_TYPE` with choices `default`, `shared` and `static`. What do you think?~ A user can specify `HDF5_USE_STATIC_LIBRARIES` themselves to override it.


cmake/thirdParties.cmake
```diff
@@ -1,23 +1,67 @@
 if(MATIO_WITH_HDF5)
-    find_package(HDF5 1.8)
+    if(NOT DEFINED HDF5_USE_STATIC_LIBRARIES)
+        if(MATIO_SHARED)
+            set(HDF5_USE_STATIC_LIBRARIES FALSE)
```
tbeu, line 4:
```suggestion
            set(HDF5_USE_STATIC_LIBRARIES FALSE)
            add_definitions(-DH5_BUILT_AS_DYNAMIC_LIB)
```

I only get it correctly linked with shared hdf5.lib on VS if H5_BUILT_AS_DYNAMIC_LIB is defined when building matio.

tbeu:
Thanks again. I was constantly failing in configuring the conan recipe to find the proper hdf5 dependency. With these fixes (and my above change setting H5_BUILT_AS_DYNAMIC_LIB as pre-processor define it is finally working, no matter if shared/static matio links shared/static hdf or whatsoever.

cmake/thirdParties.cmake
tbeu: Hm, still struggling.

MaartenBent: Looks like I made a mistake and it now always links with `hdf5::hdf5-static`. The `hdf5::hdf5-shared` target should automatically define `H5_BUILT_AS_DYNAMIC_LIB`.

What HDF5 version do you use, because I cannot find `H5_BUILT_AS_DYNAMIC_LIB` in the HDF5 1.8 CMake config file, so we have do define it for this version ourselves.

tbeu: I use hdf5/1.12.0 when testing the conan recipe.

tbeu:
What I obsvere in VS is that `-DH5_BUILT_AS_DYNAMIC_LIB` is defined instead of `H5_BUILT_AS_DYNAMIC_LIB`.

image

MaartenBent:
Have you tried a clean build? For me it is just `H5_BUILT_AS_DYNAMIC_LIB`.

tbeu:
> Have you tried a clean build? For me it is just `H5_BUILT_AS_DYNAMIC_LIB`.

Yes, deleted all and started a fresh build with same result. CMake is 3.18.2 and Conan is 1.28.1.

MaartenBent:
Maybe it is an issue with Conan. Can you add the following at line 54 in `thirdParties.cmake` to see what definitions you get from the HDF5 library:
```cmake
get_target_property(definitions hdf5::hdf5-shared INTERFACE_COMPILE_DEFINITIONS)
message("${definitions}")
```
For me it prints: `H5_BUILT_AS_DYNAMIC_LIB`

tbeu:
It always uses the else branch:

```cmake
        # results from CMake FindHDF5
        set_target_properties(HDF5::HDF5 PROPERTIES
            INTERFACE_INCLUDE_DIRECTORIES "${HDF5_INCLUDE_DIRS}"
            INTERFACE_LINK_LIBRARIES "${HDF5_LIBRARIES}"
            INTERFACE_COMPILE_DEFINITIONS "${HDF5_DEFINITIONS}"
        )
```

MaartenBent:
Indeed, it does not seem to have the cmake config files. But I could reproduce it and hopefully fix it.

tbeu:
> Indeed, it does not seem to have the cmake config files. But I could reproduce it and hopefully fix it.

Conformed. It is working now. Thanks a lot!

tbeu (approved):
Excellent work. Thank you.

tbeu:
Argh, too early.

There is a warning in https://travis-ci.org/github/tbeu/matio/jobs/724596042

```
CMake Warning:

  Manually-specified variables were not used by the project:

    ZLIB_ROOT
```

and later when testing

```
+./tools/matdump -d ./MSL/Modelica/Resources/Data/Tables/test_v7.mat s
Compressed variable found in "./MSL/Modelica/Resources/Data/Tables/test_v7.mat", but matio was built without zlib support
An error occurred in reading the MAT file
Couldn't find variable s in the MAT file
```

Thus, it does not seem to correctly build with zlib if used from hdf5.

PR #153, opened 2020-09-14
Add shared object versioning to CMake

tbeu:
@MaartenBent I am not sure if this is necessary, but we have the equivalent shared object versioning also in https://github.com/tbeu/matio/blob/bebd29124b6499fa77c672a0ffe08591c04ded41/src/Makefile.am#L29-L34 when using GNU Autotools.

The CMake commands were kindly taken from https://github.com/c-ares/c-ares.

### DIFF
--- a/cmake/src.cmake
+++ b/cmake/src.cmake
@@ -42,37 +42,52 @@ if(NOT MSVC)
 endif()
 
 if(MATIO_SHARED)
-    add_library(matio SHARED ${MATIO_SOURCES})
+    add_library(${PROJECT_NAME} SHARED ${MATIO_SOURCES})
 else()
-    add_library(matio STATIC ${MATIO_SOURCES})
+    add_library(${PROJECT_NAME} STATIC ${MATIO_SOURCES})
 endif()
-add_library(matio::matio ALIAS matio)
+add_library(${PROJECT_NAME}::${PROJECT_NAME} ALIAS ${PROJECT_NAME})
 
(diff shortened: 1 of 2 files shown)


MaartenBent (approved):
I'm not familiar with ` -version-info` in either autotools or CMake, so I don't have much to comment. The CMake code itself looks fine.

cmake/src.cmake
```diff
@@ -82,10 +97,10 @@ set(MATIO_PUBLIC_HEADERS
 if(STDINT_MSVC)
     set(MATIO_PUBLIC_HEADERS ${MATIO_PUBLIC_HEADERS} ${PROJECT_SOURCE_DIR}/visual_studio/stdint_msvc.h)
 endif()
-set_target_properties(matio PROPERTIES PUBLIC_HEADER "${MATIO_PUBLIC_HEADERS}")
+set_target_properties(${PROJECT_NAME} PROPERTIES PUBLIC_HEADER "${MATIO_PUBLIC_HEADERS}")
 
 # 'make install' to the correct locations (provided by GNUInstallDirs).
-install(TARGETS matio EXPORT libmatio
+install(TARGETS ${PROJECT_NAME} EXPORT libmatio
```
MaartenBent, line 103: You could change this to `lib${PROJECT_NAME}` as well, just like line 64.

tbeu: Done. Thanks.

PR #155, opened 2020-09-17
Don't create HDF5::HDF5 target if it already exists + require zlib/hdf5 if enabled

madebr:
Hey,

These changes should make matio build on cci without any patches.
(apart from the fix to the linker flags)

By requiring the dependencies, you are 100% sure that matio has them enabled.
Because it might always not be found.

Kind regards


MaartenBent (changes requested):
Checking if `TARGET HDF5::HDF5` does not exist, is fine with me.

But adding `REQUIRED` results in failures when using HDF5 packages downloaded from the HDF5 website. The config files from HDF5 check for an exact version, so 1.10 or 1.12 is not found. You can see this in the Travis CI failed build https://travis-ci.org/github/tbeu/matio/jobs/728159848

Also, matio can still be built when a library is not found, so it is not necessary to fail. This is what you suggested in https://github.com/tbeu/matio/pull/150#pullrequestreview-480122597 (collapsed comment).

tbeu (changes requested):
For the sake of build system consistency: I'd actually prefer to build Hdf5 with CMake if Matio is built by CMake and use Autotools consistently otherwise.

madebr:
> For the sake of build system consistency: I'd actually prefer to build Hdf5 with CMake if Matio is built by CMake and use Autotools consistently otherwise.

Ok, but why is zlib built with autotools instead of cmake? Isn't requiring this an arbitrary requirement?

tbeu:
> > For the sake of build system consistency: I'd actually prefer to build Hdf5 with CMake if Matio is built by CMake and use Autotools consistently otherwise.
> 
> Ok, but why is zlib built with autotools instead of cmake? Isn't requiring this an arbitrary requirement?

Right. In that casse, zlib should be built by CMake, too. Or use system-wide zlib.

MaartenBent:
zlib is not build with autotools in the Travis script, it is/was enclosed with `if [[ "${USE_CMAKE:-no}" == "no" ]]`.

It is built together with the HDF5 libraries when executing `ctest`.

And in this case `find_package(ZLIB)` is not used at all, since the HDF5 config file defines the `hdf5::zlib-static` or `hdf5::zlib-shared` target.

madebr:
> zlib is not build with autotools in the Travis script, it is/was enclosed with `if [[ "${USE_CMAKE:-no}" == "no" ]]`.
> 
> It is built together with the HDF5 libraries when executing `ctest`.
> 
> And in this case `find_package(ZLIB)` is not used at all, since the HDF5 config file defines the `hdf5::zlib-static` or `hdf5::zlib-shared` target.

Correct.
I wasn't expecting this complicated configuration because I am used to always use system libraries (or conan system packages).
Personally, I try to avoid private zlib libraries because this introduces duplicate symbols.
But it's find for ci :smile: .

Anyhow, I reverted the hdf5-cmake-ci change.

.ci/travis_before_script.sh
```diff
@@ -24,7 +24,7 @@ if [[ "${USE_CMAKE:-no}" == "yes" ]] &&  [[ "${USE_CONAN:-no}" == "no" ]]; then
     mkdir -p $BUILD_DIR
     pushd $BUILD_DIR
     cmake $SRC_DIR -DCMAKE_BUILD_TYPE=Debug \
-                   -DCMAKE_PREFIX_PATH=$HDF5_DIR/share/cmake \
+                   -DHDF5_ROOT=$HDF5_DIR \
```
MaartenBent, line 27: Can you revert this as well? So CI uses the targets specified by `hdf5-config`.

madebr: I think HDF5 uses hdf5-config.
See the latest item at https://cmake.org/cmake/help/latest/module/FindHDF5.html

MaartenBent: Yes, but it needs to know where to find `hdf5-config`. That is wat `CMAKE_PREFIX_PATH=$HDF5_DIR/share/cmake` is for. Now it just searches for `libhdf5.so` in `HDF5_ROOT`.

You can also see the difference in travis output. With `hdf5-config`:
```
--- Found HDF5: hdf5::hdf5-shared (found version "1.10.6")
```
Without:
```
-- Found HDF5: /home/travis/build/tbeu/matio/hdf5-1.10.7/hdf5/lib/libhdf5.so (found suitable version "1.10.7", minimum required is "1.8")  
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found suitable version "1.2.8", minimum required is "1.2.3") 
```

MaartenBent (changes requested):
I'd still like to see the `travis_before_script.sh` change be reverted, so it uses `hdf5-config.cmake` instead of the system libraries.

I'll leave the decision about he `REQUIRED` change to you @tbeu . If hdf5/zlib cannot be found, should it abort with an error, or continue building without hdf5/zlib support?

.ci/travis_before_script.sh
madebr, line 27:
```suggestion
                   -DCMAKE_PREFIX_PATH=$HDF5_DIR/share/cmake \
```

MaartenBent:
I'm getting this same error now as well, because `FindHDF5` in CMake 3.19 defines a `HDF5::HDF5` target. I propose a different solution, see #159.

The other change in this PR (adding `REQUIRED`) is also still open for discussion. Since the CMake project tries to mirror the behaviour of configure, I suggest to do the same here as well: if hdf5 or zlib is not found, continue without supporting it (i.e don't add `REQUIRED`).

tbeu:
@MaartenBent Not being a CMake expert myself, I think I'll go with your alternate PR #159.

> The other change in this PR (adding REQUIRED) is also still open for discussion. Since the CMake project tries to mirror the behaviour of configure, I suggest to do the same here as well: if hdf5 or zlib is not found, continue without supporting it (i.e don't add REQUIRED).

Yes, sounds reasonable.

PR #159, opened 2020-11-03
Don't use existing CMake target names.

MaartenBent:
When using CMake 3.19.0-rc2 I get the same error that #155 tries to solve:

```
CMake Error at cmake/thirdParties.cmake:34 (add_library):
add_library cannot create imported target "HDF5::HDF5" because another target with the same name already exists.
```

I propose a different solution. Use the `MATIO` namespace for imported libraries. Also check and use the `HDF5::HDF5` and `ZLIB::ZLIB` targets if they exist.

### DIFF
--- a/cmake/src.cmake
+++ b/cmake/src.cmake
@@ -65,11 +65,11 @@ if(WIN32)
 endif()
 
 if(HDF5_FOUND)
-    target_link_libraries(${PROJECT_NAME} PUBLIC HDF5::HDF5)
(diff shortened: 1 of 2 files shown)


tbeu:
@madebr Does this also work for you?

madebr:
I'm fine with that.
I'll close my pr.

tbeu (approved):
Thanks. Going to merge.

PR #161, opened 2020-12-20
Fix double lib prefix in CMake MinGW builds.

MaartenBent:


MaartenBent:
There seems to be some problem with fetching the submodule from sourceforge causing some builds to fail. But the macOS build using homebrew is fixed.

tbeu (approved):
Thanks.

PR #170, opened 2021-05-31
Add new arch loongarch64

loongson-zjl:
Add new arch loongarch64.
Signed-off-by: zhangjialing <zhangjialing@loongson.cn>


tbeu:
Thanks for the PR. You might want to add your CPU also to https://sourceforge.net/p/predef/wiki/Architectures/.

Issue #172, opened 2021-06-03
macOS cmake build fails with `ld: unknown option: --no-undefined` with dynamic libs


seanm:
```
mkdir matio-bin
cd matio-bin
cmake -DMATIO_MAT73=OFF -DMATIO_WITH_HDF5=OFF ../matio
make
```

results in:

```
[ 77%] Linking C shared library libmatio.dylib
ld: unknown option: --no-undefined
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [libmatio.11.0.3.dylib] Error 1
make[1]: *** [CMakeFiles/matio.dir/all] Error 2
make: *** [all] Error 2
```

No such problem when using `MATIO_SHARED=OFF`.

This is with Xcode 11.3.1 on macOS 10.14.

MaartenBent:
Looks like LLD (LLVM Linker) on macOS (maybe on all platforms) does not support these options. Suitable replacements seem:
`-Wl,-undefined,error`
`-Wl,-exported_symbols_list,${PROJECT_SOURCE_DIR}/src/matio.sym`

The second one does cause a new problem, it expects all definitions in `matio.sym` to start with an `_`. GNU linker seems to handle this without problem, so I think it is fine to add them. Otherwise a second `.sym` file could be used.

I'll create a PR to fix this in the CMake build system.

PR #174, opened 2021-06-06
Fix CMake shared linker options on macOS

MaartenBent:
See #172

The linker on macOS does not recognize the used linker options, so use different options that should have the same effect.
The `exported_symbols_list` option does not work with the names in `matio.sym`, it requires symbol names starting with `_`, so create a copy of `matio.sym` and modify it when necessary.
Symbol names starting with `_` also works with the GNU linker but not with the LLVM Linker on Linux, so don't modify the original file.


MaartenBent:
Error seems not related to my CMake changes, but something related to conan/python (maybe it requires python3?).

tbeu:
Thank you for the PR. Should we have a macos+CMake build at Travis CI, too?

tbeu:
> Error seems not related to my CMake changes, but something related to conan/python (maybe it requires python3?).

Yes, I already noticed and thought it is related to the Conan cache. But Py3 is another useful hint.

tbeu:
> Sure, let me add it.

Thank you. Very diligent work.

PR #182, opened 2021-11-18
Update conan-center index url and linker checks for CMake

tbeu:
When using a VS with non-English compiler output, the `check_linker_flags` of CMake try-compile does not recognize the linker warnings. When building matio later I then get the two wanings

```
LINK : warning LNK4044: Nicht erkannte Option /Wl,--no-undefined; wird ignoriert.
LINK : warning LNK4044: Nicht erkannte Option /Wl,--retain-symbols-file,C:/Projekte/matio/win64/src/matio.sym; wird ignoriert.
```

Since there seems no general way to enable non-English patterns my only remedy was to avoid the `check_linker_flags` for VS.


MaartenBent (approved):
CMake change look good to me. These checks are not needed for MSVC.

PR #188, opened 2022-03-21
Add LFS

### LINKED ISSUE #157: Mat_GetDir takes an infinite time to return values
tested on windows matio 1.5.18
[test_suites.zip](https://github.com/tbeu/matio/files/5284857/test_suites.zip)

```
  char** variableNames = Mat_GetDir(matfile, &nVars);
```
mat contains a struct 'B' with many others struct as field.

### LINKED ISSUE #176: Library will not read files greater than 2Gb in size
The internal data structure members that store file positions (_mat_t.bof, matvar_internal.datapos) are 'long' data types which are inconsistent among platforms.  By conditionally making these 64-bit integers we can support files greater than 2Gb on largefile-capable 32-bit systems as well as 64-bit Windows.  This will also necessitate using 64-bit versions of fseek() and ftell() where available.

I have encountered Matlab Level 5 formatted files greater than 2Gb as output from the SWAN numerical model (http://swanmodel.sourceforge.net/).  Matlab will read these just fine, and the Matlab file format specification does not seem to prohibit large files.

I have attached a patch against matio version 1.5.19 that I have tested working on Windows 32- and 64-bit using Visual Studio and it should also work on most Unix/Linux systems, although untested.  The approach relies on exposing 64-bit versions of fseeko() and ftello() in stdio.h but the method used should really be implemented as a compiler flag in the build system and/or in the generation of matioConfig.h, so mine is not the most elegant implementation.

Nevertheless, it works and my patch contribution is attached: 

  [largefile.zip](https://github.com/tbeu/matio/files/6848610/largefile.zip)

Don.

tbeu:
Fix #176.

Will squash commits before merge.

@MaartenBent Please check CMake config.


MaartenBent:
Looks good, this is what I had in mind for CMake too.

Just one remark about `_LARGEFILE64_SOURCE`. Shouldn't it be `#cmakedefine _LARGEFILE64_SOURCE 1`.
Or maybe even `#cmakedefine _LARGEFILE64_SOURCE @_LARGEFILE64_SOURCE@`?

PR #192, opened 2022-09-27
CMake: fix version comparison

nim65s:
Hi again,

On CentOS 7, with CMake 3.7.2, I get the following issue:
```
CMake Error at cmake/compilerOptions.cmake:93 (include):
include could not find load file:

CheckLinkerFlag
CMakeLists.txt:30 (include)


CMake Error at cmake/compilerOptions.cmake:94 (check_linker_flag):
Unknown CMake command "check_linker_flag".
CMakeLists.txt:30 (include)
```


MaartenBent (approved):
This is indeed the correct way to compare versions.

tbeu:
Thanks for reporting and fixing.

PR #195, opened 2022-10-26
fix autoconf getopt_long check

niclasr:
NULL is defined in stddef.h according to the C11 standard, so include it if it exists. opt is missing its data type so compilation of the code always fails. Fix it by adding int, the return type of getopt_long.


tbeu:
Thanks for the fix. Very appreciated.

PR #197, opened 2023-01-13
Introduce predicated iteration functions

Ri0n:
Reading a Matio variable from an HDF5 files sometimes can be too slow. Besides it's often necessary to iterate over all variable to read just one. This takes a lot of time. This change allows to skip unnecessary readings

### DIFF
--- a/src/mat.c
+++ b/src/mat.c
@@ -2576,6 +2576,24 @@ Mat_VarReadDataLinear(mat_t *mat, matvar_t *matvar, void *data, int start, int s
 matvar_t *
 Mat_VarReadNextInfo(mat_t *mat)
 {
+    return Mat_VarReadNextInfoPredicate(mat, NULL, NULL);
+}
+
+/** @brief Reads the information of the next variable in a filtered MAT file
+ *
+ * Reads the next variable's information (class,flags-complex/global/logical,
+ * rank,dimensions, name, etc) from the Matlab MAT file. Calls a user callback
+ * to check where the variable has to be fully read of skipped.
+ * If skipped tries to read next till accepted or EOF.
+ * After reading, the MAT file is positioned past the current variable.
+ * @ingroup MAT
+ * @param mat Pointer to the MAT file
+ * @return Pointer to the @ref matvar_t structure containing the MAT
+ * variable information
+ */
+matvar_t *
+Mat_VarReadNextInfoPredicate(mat_t *mat, mat_iter_pred_t pred, const void *user_data)
+{
     matvar_t *matvar;
     if ( mat == NULL )
         return NULL;
@@ -2586,7 +2604,7 @@ Mat_VarReadNextInfo(mat_t *mat)
             break;
         case MAT_FT_MAT73:
 #if defined(MAT73) && MAT73
-            matvar = Mat_VarReadNextInfo73(mat);
+            matvar = Mat_VarReadNextInfo73(mat, pred, user_data);
 #else
             matvar = NULL;
 #endif
@@ -2600,6 +2618,13 @@ Mat_VarReadNextInfo(mat_t *mat)
     }
 
     return matvar;
+}
+
+static int
+Mat_IteratorNameAcceptor(const char *name, const void *user_data)
+{
+    const char *required_name = user_data;
+    return name && required_name && strcmp(name, required_name) == 0;
 }
 
 /** @brief Reads the information of a variable with the given name from a MAT file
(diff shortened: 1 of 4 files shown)


tbeu:
@Ri0n Can you please force-push again to trigger the Travis CI pipeline. Thanks.

src/matio.h
```diff
@@ -327,7 +333,9 @@ EXTERN int Mat_VarReadDataLinear(mat_t *mat, matvar_t *matvar, void *data, int s
                                  int edge);
 EXTERN matvar_t *Mat_VarReadInfo(mat_t *mat, const char *name);
 EXTERN matvar_t *Mat_VarReadNext(mat_t *mat);
+EXTERN matvar_t *Mat_VarReadNextPredicate(mat_t *mat, mat_iter_pred_t pred, const void *user_data);
 EXTERN matvar_t *Mat_VarReadNextInfo(mat_t *mat);
+EXTERN matvar_t *Mat_VarReadNextInfoPredicate(mat_t *mat, mat_iter_pred_t pred, const void *user_data);
```
tbeu, line 338: Is it really necessary to change the public API of matio if only relevent for MAT 7.3?

src/mat73.c
```diff
@@ -3173,16 +3178,19 @@ Mat_VarReadNextInfoIterate(hid_t id, const char *name, const H5L_info_t *info, v
 
     /* FIXME: follow symlinks, datatypes? */
 
+    mat_data = (struct ReadNextIterData *)op_data;
+
     /* Check that this is not the /#refs# or /"#subsystem#" group */
     if ( 0 == strcmp(name, "#refs#") || 0 == strcmp(name, "#subsystem#") )
         return 0;
+    if (mat_data && mat_data->pred && mat_data->pred(name, mat_data->pred_user_data) == 0)  // do we need to skip it?
```
tbeu, line 3186: Please use a C89 style comment.

```suggestion
    if (mat_data && mat_data->pred && mat_data->pred(name, mat_data->pred_user_data) == 0)  /* do we need to skip it? */
```

src/mat.c
```diff
@@ -2722,27 +2751,50 @@ Mat_VarRead(mat_t *mat, const char *name)
...
+                /* Reset the file position */
+                (void)fseeko((FILE *)mat->fp, fpos, SEEK_SET);
+            }
+            break;
         }
-    } else if ( mat->version != MAT_FT_MAT73 ) {
-        /* Reset the file position */
-        (void)fseeko((FILE *)mat->fp, fpos, SEEK_SET);
-    }
+    } while (pred && pred(matvar->name, user_data) == 0);  // for 7.3 the predicate will be called one extra time
```
tbeu, line 2797: Please use a C89 style comment.

```suggestion
    } while (pred && pred(matvar->name, user_data) == 0);  /* for 7.3 the predicate will be called one extra time */
```

tbeu (commented):
Thanks for the contribution! Do you have any benchmarks demonstrating the performance gain?

src/matio.h
Ri0n: `Mat_VarReadNextPredicate` was required in a project where we have matio integrated.
Not the case for `Mat_VarReadNextInfoPredicate` though. I can hide it.

tbeu: Should be symmetric. So I am OK to keep as is.

tbeu:
If you want to you can add yourself as contributor to section 1.2 of both files README.md and README.

Ri0n:
Wrt performance.
There was a problem with one file. And this PR is an alternative fix to one merged in  https://github.com/tbeu/matio/pull/198
In both cases file loading time was reduced from 2 minutes to 5-7 seconds.

The crux of the problem is in long string string members in `header` variable. 
IIRC 5+ strings. The strings themselves were quite short but declared at 1000+ symbols.
Each symbol was resolved as a variable with `H5Iget_name` somewhere on the way, making all the reading extremely slow.

One way to solve this was to read `header` directly with HDF5 and use this PR's fix to read everything else but `header` with help of matio.  And basically it's what currently applied in our project.

tbeu:
Thanks for the numbers and update.

Do you suggest instead to not merge this PR after #198 got merged?

Ri0n:
I think both together are good to raise performance to its possible maximum.

tbeu:
Argh, I missed that #198 failed the test suite in 18 tests, see https://app.travis-ci.com/github/tbeu/matio/jobs/593339008#L5197

Unfortunately, this is hidden and not marked as failure by Travis CI.

Ri0n:
> Argh, I missed that https://github.com/tbeu/matio/pull/198 failed the test suite in 18 tests, see https://app.travis-ci.com/github/tbeu/matio/jobs/593339008#L5197

That what I was afraid of. It's touches too many places.

tbeu:
Never mind. Do you think you are able to fix the tests?

Ri0n:
Probably I can fix it with some guidance. Unfortunately it's no more a priority task at my job since an alternate fix exists. So I need to invest my own time which I'm always lack of.
In any case I can do nothing till the end of the week.

tbeu (changes requested):
Testsuite is failing by 18 tests (same as #198).

tbeu:
CI succeeds now.

PR #198, opened 2023-01-13
Don't use slow H5Iget_name

Ri0n:
we had files with approximately same data in mat5 and mat73/hdf5 format.
reading of mat73 file was many times slower than mat5. It turned out Matio calls H5Iget_name way too often, and H5Iget_name is a very resource-expensive function.

See similar issue https://forum.hdfgroup.org/t/h5i-get-name-call-is-very-slow-for-hdf5-file-5-gb/8246


tbeu:
@Ri0n Can you please force-push again to trigger the Travis CI pipeline. Thanks.

tbeu (approved):
Thank you for the contibution.

tbeu:
I needed to revert the changes by 445908815e161968e40435d60a095d560f6a88e1 to make the testsuite passing again,

Ri0n:
sure. I had no time to take a look at this myself. but the second patch I provided we currently use in production without any issue.

PR #210, opened 2023-08-08
Fix Heap-buffer-overflow WRITE in H5MM_memcpy

sashashura:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58262

The root cause is that `H5Sget_simple_extent_dims` doesn't fail and doesn't initialize `*nfields` for scalar types, so it is left zero. Behavior of `calloc` call with zero size is implementation dependent, but in this case it returns a non null memory which is valid to use only with `free`.

### DIFF
--- a/src/mat73.c
+++ b/src/mat73.c
@@ -644,10 +644,10 @@ Mat_H5ReadFieldNames(matvar_t *matvar, hid_t dset_id, hsize_t *nfields)
         H5Aclose(attr_id);
         return MATIO_E_GENERIC_READ_ERROR;
     } else {
+        if ( err == 0 ) {
+            *nfields = 1;
+        }
         err = MATIO_E_NO_ERROR;
-    }
-    if ( err == 0 ) {
-        *nfields = 1;
     }
     fieldnames_vl = (hvl_t *)calloc((size_t)(*nfields), sizeof(*fieldnames_vl));
     if ( fieldnames_vl == NULL ) {
--- a/src/mat73.c
+++ b/src/mat73.c
@@ -646,6 +646,8 @@ Mat_H5ReadFieldNames(matvar_t *matvar, hid_t dset_id, hsize_t *nfields)
     } else {
         err = MATIO_E_NO_ERROR;
     }
+    if (*nfields == 0)
+        *nfields = 1;
     fieldnames_vl = (hvl_t *)calloc((size_t)(*nfields), sizeof(*fieldnames_vl));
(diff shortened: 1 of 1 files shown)


src/mat73.c
```diff
@@ -651,6 +651,8 @@ Mat_H5ReadFieldNames(matvar_t *matvar, hid_t dset_id, hsize_t *nfields)
     } else {
         err = MATIO_E_NO_ERROR;
     }
+    if (*nfields == 0)
+        *nfields = 1;
```
tbeu, line 655:
```suggestion
    if ( *nfields == NULL ) {
        *nfields = 1;
    }
```

Thanks for the fix. Please do not forget to run clang-format.

src/mat73.c
```diff
@@ -651,6 +651,9 @@ Mat_H5ReadFieldNames(matvar_t *matvar, hid_t dset_id, hsize_t *nfields)
     } else {
         err = MATIO_E_NO_ERROR;
     }
+    if ( *nfields == 0 ) {
+        *nfields = 1;
```
tbeu, line 655: I actually wonder if we should return with error (MATIO_E_OPERATION_NOT_SUPPORTED or MATIO_E_GENERIC_READ_ERROR) instead.

tbeu:
Need to get the CI working first which is broken since merge of #198.

tbeu:
Travis CI reports Error "https://github.com/modelica-tools/msl-release" for certain test cases.

sashashura:
Do you mean the pull request broke some tests?

tbeu:
> Do you mean the pull request broke some tests?

Yes, this seems to be the case.

tbeu:
I fixed the CI and rebased your PR branch such that you can now see the crashes of the testsuite.

sashashura:
My bad, I have switched to using `err` last minute and didn't notice it is overridden few lines above.

sashashura:
Travis didn't start for some reason.

tbeu:
Travis CI credit balance is negative again.

sashashura:
But it ran for https://github.com/tbeu/matio/pull/214 which was made a bit later.
May I ask why you use two build systems - appveyor and travis, but not github actions at the same time?

sashashura:
Looks like everything has passed except 1 of 3 macos builds that timed out.

sashashura:
The https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58262 was closed!

Issue #209, opened 2023-08-04
Buliding with afl++ gives linker error `ld: unknown option: --retain-symbols-file`


seanm:
I've recently tried building my code and all its dependencies with [afl++](https://github.com/AFLplusplus/AFLplusplus) and matio failed to build with this error:

```
afl-cc++4.08a by Michal Zalewski, Laszlo Szekeres, Marc Heuse - mode: LLVM-PCGUARD
ld: unknown option: --retain-symbols-file
clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
```

Luckily, I just removed `--retain-symbols-file` and things still linked correctly for me.

Perhaps matio doesn't actually need this option? Or perhaps it could be used less often?

tbeu:
Did you use the CMake bases build scripts or the autotools/configure scripts?

seanm:
CMake. Sorry, should have said that...

tbeu:
@MaartenBent Do you have an idea why clang-16 does not support the `--retain-symbols-file` linker flag set in https://github.com/tbeu/matio/blob/1139fe9685f5ebfdd36a43a274d41a8a11a445a9/cmake/src.cmake#L42 ?

MaartenBent:
I don't think the problem is specifically clang-16, it compiles and links successfully for me on Windows with clang-16. Maybe afl++ uses their own  linker instead of lld that doesn't support this option.

We could check if `-Wl,--retain-symbols-file` is supported, like we already do for `-Wl,--no-undefined`. I created the patch below as a possible fix. I still have to test if the `-Wl,-exported_symbols_list` test passes successfully on macOS.

These options are added because [configure](https://github.com/tbeu/matio/blob/1139fe9685f5ebfdd36a43a274d41a8a11a445a9/src/Makefile.am#L50) has them too. I think they are not explicitly need, but they provide better control of what symbols are (not) exported.


```diff
 cmake/compilerOptions.cmake | 8 ++++++--
 1 file changed, 6 insertions(+), 2 deletions(-)

diff --git a/cmake/compilerOptions.cmake b/cmake/compilerOptions.cmake
index f097daf..3c6b3f9 100644
--- a/cmake/compilerOptions.cmake
+++ b/cmake/compilerOptions.cmake
@@ -92,10 +92,14 @@ if (NOT MSVC)
     if(${CMAKE_VERSION} GREATER "3.17")
         include(CheckLinkerFlag)
[...]

PR #212, opened 2023-08-10
Fix linker options

tbeu:
Closes https://github.com/tbeu/matio/issues/209

@seanm Please confirm if it works for you when using CMake &GreaterEqual; 3.17. Thanks.


MaartenBent:
The `HAVE_LINK_EXPORTED_SYMBOLS_LIST` test always seems to fail, so now those flags are never added on macOS. I haven't been able to make it pass.
I suggest removing the test and changing the check to `elseif(APPLE AND HAVE_LINK_UNDEFINED_ERROR)`.

tbeu:
> I suggest removing the test and changing the check to `elseif(APPLE AND HAVE_LINK_UNDEFINED_ERROR)`.

Done by https://github.com/tbeu/matio/pull/212/commits/400cfec4059d30bd918fb895bed9187e088a9b00.

seanm:
Although I'm using macOS, my impression is this issue is related to AFL more than it is to macOS.  I would expect the same problem with AFL on linux too...

seanm:
> Does the updated patch still work for you?

It does.

tbeu:
> > Does the updated patch still work for you?
> 
> It does.

With invalidated CMake cache, I hope.

MaartenBent:
Thanks for testing.

Maybe `--retain-symbols-file` works on Linux too when using clang and/or ldd. But I couldn't get it to work on Windows. Thats why I suggested to limit it to `APPLE` for now. Since these options were [originally](https://github.com/tbeu/matio/commit/c80718f4259654bdf40688ef3b21f92faa265e6c) added only for macOS because it didn't work with `--no-undefined`.

This should be unaffected by the CMake cache. The result variables of `check_linker_flag` are cached, but if/else is evaluated every time you rerun CMake. Though testing with a clean build directory is always good.

MaartenBent (approved):

PR #215, opened 2023-08-15
Avoid crashes with corrupted files

Ri0n:
Our customers quite often come with corrupted files which lead to crashes somewhere in Matio.
This PR fixes the crashes we discovered so far.


tbeu (approved):
Thank you! LGTM.

PR #223, opened 2023-11-17
Update version

tbeu:
This is actually two proposed changes:

1. Set SOVERSION and VERSION always (does not hurt to also have it for static/object libraries)
2. Set the VERSION to match the project version.


MaartenBent (approved):
Looks fine to me.

PR #224, opened 2023-11-19
Add CMake option to link MSVC runtime library statically

tbeu:


cmake/options.cmake
```diff
@@ -17,6 +17,22 @@ option(MATIO_SHARED "Build shared matio library, disable for static library" ON)
 # Option to enable position-independent code (PIC)
 option(MATIO_PIC "Enable position-independent code (PIC), i.e., compilation with the -fPIC flag" ON)
 
+# Option to link the MSVC runtime library statically
+option(MATIO_STATIC_LINK_CRT "Link the MSVC runtime library statically" OFF)
+
+if(POLICY CMP0091)
+    # CMake >= 3.15 has CMAKE_MSVC_RUNTIME_LIBRARY to set the MSVCC runtime library
+    if(WIN32 AND CMAKE_GENERATOR MATCHES "Visual Studio .*|NMake .*")
```
MaartenBent, line 25: I think you can just use `if(MSVC)` here.

cmake/options.cmake
```diff
@@ -17,6 +17,22 @@ option(MATIO_SHARED "Build shared matio library, disable for static library" ON)
...
+        if(MATIO_STATIC_LINK_CRT)
+            message(STATUS "Configuring to link the MSVC runtime library statically")
```
MaartenBent, line 27: Is it necessary to keep these messages?

cmake/options.cmake
```diff
@@ -17,6 +17,22 @@ option(MATIO_SHARED "Build shared matio library, disable for static library" ON)
 # Option to enable position-independent code (PIC)
 option(MATIO_PIC "Enable position-independent code (PIC), i.e., compilation with the -fPIC flag" ON)
 
+# Option to link the MSVC runtime library statically
+option(MATIO_STATIC_LINK_CRT "Link the MSVC runtime library statically" OFF)
+
+if(POLICY CMP0091)
```
MaartenBent, line 23: Maybe move the option inside the policy check, so it is not visible in old CMake GUIs.
And also inside the `MSVC` check. Or do you plan to add something similar for other compilers too?

MaartenBent (commented):
I only did a code review. I have not yet tried to build it.

cmake/options.cmake
tbeu: Change to VERBOSE

tbeu:
Because of dependencies it is hard to get it consistently right. Closing.

PR #225, opened 2023-11-21
Set RC file

tbeu:


MaartenBent:
This is not necessary. The resource file is already added in
https://github.com/tbeu/matio/blob/0884b80a7ee5783b6193d7e2c58f997d121d3245/cmake/src.cmake#L30

You can verify it works by checking the details of the .dll properties. It sets the file description, copyright, and other fields.

PR #232, opened 2024-01-31
Recognize __arm64__.

maltelenz:
macOS ARM chips define this.

### DIFF
--- a/src/mat4.c
+++ b/src/mat4.c
@@ -167,7 +167,7 @@ Mat_VarWrite4(mat_t *mat, const matvar_t *matvar)
 #elif defined(__i386__) || defined(__alpha__) || defined(__ia64) || defined(__ia64__) ||   \
     defined(_M_IX86) || defined(_M_IA64) || defined(_M_ALPHA) || defined(__amd64) ||       \
     defined(__amd64__) || defined(_M_AMD64) || defined(__x86_64) || defined(__x86_64__) || \
-    defined(_M_X64) || defined(__bfin__) || defined(__loongarch64)
+    defined(_M_X64) || defined(__bfin__) || defined(__loongarch64) || defined(__arm64__)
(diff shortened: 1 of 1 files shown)


tbeu:
Thanks for the PR. Unfortunately, I cannot find that macro in https://sourceforge.net/p/predef/wiki/Home/.

src/mat4.c
```diff
@@ -167,7 +167,7 @@ Mat_VarWrite4(mat_t *mat, const matvar_t *matvar)
 #elif defined(__i386__) || defined(__alpha__) || defined(__ia64) || defined(__ia64__) ||   \
     defined(_M_IX86) || defined(_M_IA64) || defined(_M_ALPHA) || defined(__amd64) ||       \
     defined(__amd64__) || defined(_M_AMD64) || defined(__x86_64) || defined(__x86_64__) || \
-    defined(_M_X64) || defined(__bfin__) || defined(__loongarch64)
+    defined(_M_X64) || defined(__bfin__) || defined(__loongarch64) || defined(__arm64__)
```
maltelenz, line 170:
```suggestion
    defined(_M_X64) || defined(__bfin__) || defined(__loongarch64) || defined(__aarch64__)
```

The [architechtures](https://sourceforge.net/p/predef/wiki/Architectures/history) page hasn't changed since 2015, so I guess it isn't surprising it doesn't know about Apple ARM defines. However, https://developer.apple.com/documentation/apple-silicon/addressing-architectural-differences-in-your-macos-code uses `__aarch64__` in one of its examples. This also works for me on my M1 laptop.

Lacking more explicit documentation by Apple, I guess this is a good choice?

From `echo | clang -dM -E -` I can see that the clang on the M1 provides `__aarch64__`, `__arm64` as well as `__arm64__`.

PR #237, opened 2024-02-08
HDF5-based format: update detection of HDF5 version for UTF-8-based filenames

chris-se:
As discussed in https://github.com/tbeu/matio/pull/236 here is the dependent change that updates the minimum HDF5 version to 1.10.6. The commit message contains a longer explanation for the rationale behind the change.


tbeu:
Thanks!

Issue #233, opened 2024-02-05
MinGW Build Regression


chris-se:
The fix for <https://github.com/tbeu/matio/issues/187> caused a build regression in MinGW for me (amd64, CMake Buildsystem):

 - MinGW (for compatibility w/ build systems that expect glibc) allows the user to specify `-lm` for the math functions, and automatically maps them to the correct CRT. But it does **not** provide a `-lc`, because normally nobody explicitly specifies that.
 - The fix for the OpenBSD issue caused the build system of matio to automatically add `-lc` to the compiler line due to the commit <https://github.com/tbeu/matio/commit/f058e5dfbaa8bc49fbe76631b6dbf7637be04b10>.

This causes MinGW builds to fail, since `HAVE_LIBM` is true on those systems, but `-lc` doesn't work there.

The least wrong simple fix I could come up with (without testing on OpenBSD, which I don't use) was to simply also check for `-lc` and then only link to that if it's available. The following patch works for me on Linux, macOS and Windows:

``` patch
--- a/cmake/compilerOptions.cmake       2023-11-12 13:25:19.000000000 +0100
+++ b/cmake/compilerOptions.cmake       2024-02-05 16:29:48.804984158 +0100
@@ -70,6 +70,13 @@ endforeach()
 
 include(CheckLibraryExists)
 check_library_exists(m pow "" HAVE_LIBM)
+# OpenBSD also requires -lc (See
+# <https://github.com/tbeu/matio/issues/187> for
+# details), but that doesn't exist on all platforms
+# that do have libm (most notably not MinGW), so
+# explicitly check for the existence of -lc before
+# linking against in in src.cmake.
+check_library_exists(c fopen "" HAVE_LIBC)
 
 include(CheckCSourceCompiles)
 set(TEST_CODE_DECIMAL_POINT "
--- a/cmake/src.cmake   2023-11-12 13:25:19.000000000 +0100
+++ b/cmake/src.cmake   2024-02-05 16:26:10.244019256 +0100
@@ -68,7 +68,10 @@ if(STDINT_MSVC)
 endif()
 
 if(HAVE_LIBM)
-    target_link_libraries(${PROJECT_NAME} PUBLIC m c)
+    target_link_libraries(${PROJECT_NAME} PUBLIC m)
+    if(HAVE_LIBC)
+        target_link_libraries(${PROJECT_NAME} PUBLIC c)
+    endif()
 endif()
 
 if(MSVC)
```

[...]

Issue #187, opened 2022-03-18
link errors building master on OpenBSD due to `-Wl,--no-undefined`


seanm:
Using cmake on OpenBSD 7.0.  `make VERBOSE=1` gives:

```
/usr/local/bin/cmake -E cmake_link_script CMakeFiles/matio.dir/link.txt --verbose=1
/usr/bin/cc -fPIC  -Wl,--no-undefined -Wl,--retain-symbols-file,/home/builder/external/matio-bin/src/matio.sym -shared -Wl,-soname,libmatio.so.11 -o libmatio.so.11.0.3 CMakeFiles/matio.dir/src/endian.c.o CMakeFiles/matio.dir/src/mat.c.o CMakeFiles/matio.dir/src/io.c.o CMakeFiles/matio.dir/src/inflate.c.o CMakeFiles/matio.dir/src/mat73.c.o CMakeFiles/matio.dir/src/matvar_cell.c.o CMakeFiles/matio.dir/src/matvar_struct.c.o CMakeFiles/matio.dir/src/mat4.c.o CMakeFiles/matio.dir/src/mat5.c.o CMakeFiles/matio.dir/src/snprintf.c.o CMakeFiles/matio.dir/src/read_data.c.o  -lm /usr/lib/libz.so.6.0 -Wl,-rpath-link,/usr/X11R6/lib:/usr/local/lib
ld: error: undefined symbol: malloc
```

Running that line manually *without* the `-Wl,--no-undefined` results in a successful linking.

I'm not familiar with what that is, or why it's there...

tbeu:
Thanks for reporting. Does it work when build by GNU autotools?

seanm:
Seems like yes, and I'm able to run the tests (many fail, but whatever).

tbeu:
I added FreeBSD OS by Cirrus CI and cannot reproduce: https://cirrus-ci.com/build/5653130858725376

seanm:
Did you fix it?  Or is one expected to use autotools and not CMake?

tbeu:
No, cannot fix as I do nat have a patch if I cannot repoduce.

seanm:
Looks like there is no OpenBSD option with Cirrus CI?  I reproed in a VM with a very vanilla OpenBSD installation.  That's a fair bit of work of course, which I'm not expecting you to do...

tbeu:
PR is welcome.

seanm:
[...]

PR #238, opened 2024-02-08
cmake: add a proper test to determine if explicit linking against -lc is required

chris-se:
This pull request hopefully contains the proper way to fix the `-lc` issue raised in #187, without breaking MinGW as in #233.

This was tested on Linux and Windows/MinGW, but not on OpenBSD (I don't have a system running). However, since the check is operating-system agnostic, it should hopefully work there as well.

### DIFF
--- a/cmake/compilerOptions.cmake
+++ b/cmake/compilerOptions.cmake
@@ -88,26 +88,49 @@ check_c_source_compiles("${TEST_CODE_THOUSANDS_SEP}" HAVE_STRUCT_LCONV_THOUSANDS
 
 set(USE_GNU_LINK_FLAGS 0)
 set(USE_LLVM_MACOS_LINK_FLAGS 0)
+set(REQUIRE_EXPLICIT_LIBC_LINK 0)
 if(NOT MSVC)
-    if(${CMAKE_VERSION} VERSION_GREATER "3.17")
-        include(CheckLinkerFlag)
-        check_linker_flag(C "-Wl,--no-undefined" HAVE_LINK_NO_UNDEFINED)
-        check_linker_flag(
-            C "-Wl,--retain-symbols-file,${PROJECT_SOURCE_DIR}/src/matio.sym"
-            HAVE_LINK_RETAIN_SYMBOLS_FILE
-        )
-        check_linker_flag(C "-Wl,-undefined,error" HAVE_LINK_UNDEFINED_ERROR)
+    # OpenBSD apparently requires an explicit -lc if -Wl,--no-undefined
+    # is used, but that is NOT required on other platforms, and on even
+    # other platforms (e.g. MinGW) -lc might not even exist at all.
+    # Therefore detect if -lc works and if it is required, and omit it
+    # if it doesn't.
+    # Just running check_linker_flag() doesn't help here, because we
+    # actually need to reference a symbol from libc to cause an error.
+    # Use -shared in CMAKE_REQUIRED_LINK_OPTIONS because how symbols
+    # (esp. undefined ones) are resolved during linking can differ
+    # between shared libraries and executables.
+    set(TEST_SRC_LINK_NO_UNDEFINED "
+#include <stdlib.h>
+int main() { int* foo = (int*) malloc(sizeof(int)); free(foo); return 0; }
+")
 
-        if(HAVE_LINK_NO_UNDEFINED AND HAVE_LINK_RETAIN_SYMBOLS_FILE)
-            set(USE_GNU_LINK_FLAGS 1)
-        elseif(APPLE AND HAVE_LINK_UNDEFINED_ERROR)
-            set(USE_LLVM_MACOS_LINK_FLAGS 1)
-        endif()
-    else()
-        if(APPLE AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-            set(USE_LLVM_MACOS_LINK_FLAGS 1)
-        else()
-            set(USE_GNU_LINK_FLAGS 1)
-        endif()
+    set(CMAKE_REQUIRED_FLAGS "-fPIC")
+    set(CMAKE_REQUIRED_LINK_OPTIONS "-shared;-Wl,--no-undefined")
+    check_c_source_compiles("${TEST_SRC_LINK_NO_UNDEFINED}" HAVE_LINK_NO_UNDEFINED_IMPLICIT_LIBC)
+    set(CMAKE_REQUIRED_LINK_OPTIONS "-shared;-Wl,--no-undefined;-lc")
+    check_c_source_compiles("${TEST_SRC_LINK_NO_UNDEFINED}" HAVE_LINK_NO_UNDEFINED_EXPLICIT_LIBC)
+    # Reuse the same source code for other linker flag tests (This was
+    # previously guarded by CMake version >= 3.17 using
+    # check_linker_flag, but this variant should work for even older
+    # CMake versions.)
+    set(CMAKE_REQUIRED_LINK_OPTIONS "-shared;-Wl,--retain-symbols-file,${PROJECT_SOURCE_DIR}/src/matio.sym")
+    check_c_source_compiles("${TEST_SRC_LINK_NO_UNDEFINED}" HAVE_LINK_RETAIN_SYMBOLS_FILE)
+    set(CMAKE_REQUIRED_LINK_OPTIONS "-shared;-Wl,-undefined,error")
+    check_c_source_compiles("${TEST_SRC_LINK_NO_UNDEFINED}" HAVE_LINK_UNDEFINED_ERROR)
+    set(CMAKE_REQUIRED_FLAGS "")
+    set(CMAKE_REQUIRED_LINK_OPTIONS "")
+
+    if((HAVE_LINK_NO_UNDEFINED_IMPLICIT_LIBC OR HAVE_LINK_NO_UNDEFINED_EXPLICIT_LIBC) AND HAVE_LINK_RETAIN_SYMBOLS_FILE)
+        message(VERBOSE "Using GNU-style linker flags")
+        set(USE_GNU_LINK_FLAGS 1)
+    elseif(APPLE AND HAVE_LINK_UNDEFINED_ERROR)
+        message(VERBOSE "Using Apple-clang-style linker flags")
+        set(USE_LLVM_MACOS_LINK_FLAGS 1)
+    endif()
+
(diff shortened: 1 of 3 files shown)


chris-se:
Note: the windows-mingw test in this PR fails because the test execution doesn't work correctly, probably a minor CI setup issue. (See <https://github.com/tbeu/matio/actions/runs/7831127379/job/21366733842?pr=238>) Compare this to a different PR where the MinGW build wasn't yet fixed, and it failed already at the build stage in the windows-mingw test. (See <https://github.com/tbeu/matio/actions/runs/7827524982/job/21355503568?pr=237>)

cmake/compilerOptions.cmake
```diff
@@ -111,3 +111,39 @@ if(NOT MSVC)
...
+    # is used, but that is NOT required on other platforms, and on even
+    # other platforms (e.g. MinGW) -lc might not even exist at all.
+    # Therefore detect if -lc works and if it is required, and omit it
+    # if it doesn't.
+    # Just running check_linker_flag() doesn't help here, because we
+    # actually need to reference a symbol from libc to cause an error.
+    # Use -shared in CMAKE_REQUIRED_LINK_OPTIONS because how symbols
+    # (esp. undefined ones) are resolved during linking can differ
+    # between shared libraries and executables.
+    set(TEST_LINK_NO_UNDEFINED_EXIMPLICIT_LIBC "
```
tbeu, line 127: Is HAVE_LINK_NO_UNDEFINED some lines avove still necessary?

tbeu:
@seanm Can you please cross-check in OpenBSD? Thanks!

tbeu:
> Note: the windows-mingw test in this PR fails because the test execution doesn't work correctly, probably a minor CI setup issue. (See https://github.com/tbeu/matio/actions/runs/7831127379/job/21366733842?pr=238) Compare this to a different PR where the MinGW build wasn't yet fixed, and it failed already at the build stage in the windows-mingw test. (See https://github.com/tbeu/matio/actions/runs/7827524982/job/21355503568?pr=237)

Yes, my bad. Can ypu fix it on the fly. $MATDUMP is just different.

cmake/compilerOptions.cmake
chris-se: Potentially no, but then you'd have to restructure your CMake file quite a bit. (Especially since you have the version check for CMake in there, and otherwise always set `USE_GNU_LINK_FLAGS` to 1 if it's not an apple system.) I wanted to keep things as simple as possible to just fix the regression.

tbeu:
There now is a OpenBSD CI available. Feel free to rebase.

seanm:
> There now is a OpenBSD CI available. Feel free to rebase.

Oh, nice!  I didn't think github had that. Where can I RTFM on that?

tbeu:
What excatly? [OpenBSD CI](https://github.com/vmactions/openbsd-vm) or [rebase button](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/keeping-your-pull-request-in-sync-with-the-base-branch)?

MaartenBent (changes requested):
Looks good. But I would prefer `message(VERBOSE ...)`.
This status messages should not be relevant for end users.

CMake also has some macros to handle the `CMAKE_REQUIRED_` variables, that you could use instead of manually resetting them: https://cmake.org/cmake/help/latest/module/CMakePushCheckState.html

cmake/compilerOptions.cmake
chris-se: Actually, with my new update (see below), I've now reworked the logic as you suggested.

chris-se:
I've now updated this PR to include all of the previous feedback, including using `VERBOSE` for specific messages, as well as reworking some other things. (+ rebasing on current master)

The OpenBSD CI has now run successfully, and the CMake checks work as intended:
```
  -- Performing Test HAVE_LINK_NO_UNDEFINED_IMPLICIT_LIBC
  -- Performing Test HAVE_LINK_NO_UNDEFINED_IMPLICIT_LIBC - Failed
  -- Performing Test HAVE_LINK_NO_UNDEFINED_EXPLICIT_LIBC
  -- Performing Test HAVE_LINK_NO_UNDEFINED_EXPLICIT_LIBC - Success
```
This means that `-lc` is added on OpenBSD because that is needed there, but it's not added on other platforms, where it either isn't required (Linux) or even breaks things (MinGW).

MaartenBent (approved):

tbeu:
Thanks for fixing! Very appreciated.

tbeu:
I've just noticed that the merge breaks the linking of matdump.exe on my Cygwin setup.

The relevant configure output is:

```
-- Performing Test HAVE_LINK_NO_UNDEFINED_IMPLICIT_LIBC
-- Performing Test HAVE_LINK_NO_UNDEFINED_IMPLICIT_LIBC - Success
-- Performing Test HAVE_LINK_NO_UNDEFINED_EXPLICIT_LIBC
-- Performing Test HAVE_LINK_NO_UNDEFINED_EXPLICIT_LIBC - Success
-- Performing Test HAVE_LINK_RETAIN_SYMBOLS_FILE
-- Performing Test HAVE_LINK_RETAIN_SYMBOLS_FILE - Success
-- Performing Test HAVE_LINK_UNDEFINED_ERROR
-- Performing Test HAVE_LINK_UNDEFINED_ERROR - Success
-- Using GNU-style linker flags
```

The link command is

```
/usr/bin/cc -O3 -DNDEBUG -Wl,--enable-auto-import CMakeFiles/matdump.dir/tools/matdump.c.o CMakeFiles/matdump.dir/snprintf/snprintf.c.o -o matdump.exe -Wl,--out-implib,libmatdump.dll.a -Wl,--major-image-version,0,--minor-image-version,0  libmatio.a -lm "/cygdrive/c/Program Files/HDF_Group/HDF5/1.14.3/lib/libhdf5.lib" "/cygdrive/c/Program Files/HDF_Group/HDF5/1.14.3/lib/libszaec.lib" "/cygdrive/c/Program Files/HDF_Group/HDF5/1.14.3/lib/libaec.lib" "/cygdrive/c/Program Files/HDF_Group/HDF5/1.14.3/lib/libzlib.lib" -lshlwapi 
```

which results in all thousands of obsure error messages.

The previos link command was

```
/usr/bin/cc -O3 -DNDEBUG -Wl,--enable-auto-import CMakeFiles/matdump.dir/tools/matdump.c.o CMakeFiles/matdump.dir/snprintf/snprintf.c.o -o matdump.exe -Wl,--out-implib,libmatdump.dll.a -Wl,--major-image-version,0,--minor-image-version,0  libmatio.a -lm -lc "/cygdrive/c/Program Files/HDF_Group/HDF5/1.14.3/lib/libhdf5.lib" "/cygdrive/c/Program Files/HDF_Group/HDF5/1.14.3/lib/libszaec.lib" "/cygdrive/c/Program Files/HDF_Group/HDF5/1.14.3/lib/libaec.lib" "/cygdrive/c/Program Files/HDF_Group/HDF5/1.14.3/lib/libzlib.lib" -lshlwapi 
```

with `-lc` being the only difference.



whereas

MaartenBent:
So cygwin does need libc, even though the implicit check succeeds.

It can't find `printf`. Can you modify the test code to use `printf` instead of (or in addition to) `malloc` and `free`?

MaartenBent:
Also are you linking with msvc hdf5 libraries? I've never used cygwin myself, but mixing gcc and msvc is usually not a good idea.

tbeu:
I tried a lot - even forcing -lc, but it only works if the merge is reverted.

Yes, I also wondered why it links with MSVC static HDF5 libs.

chris-se:
Most of the messages for undefined symbols you posted come from linking against MSVC static HDF5 libs, which obviously don't use the same C runtime as Cygwin, which is why those fail. (What I don't get is why they succeed if you revert the patch? They shouldn't work at all.)

My guess is that `-lc` isn't actually what's causing this here, but something probably related to linking against static MSVC libraries (perhaps the header of the wrong C runtime is used during compilation?). I haven't used cygwin in more than a decade, but I can take a look. Could you describe your setup a bit more? Did you just install cygwin + the official Windows binaries of HDF5?

chris-se:
So interestingly, in the constellation you describe (HDF5 installed via the official binary installer) and building from within Cygwin, I get the following error already compiling in cygwin **before** the CMake change:

```
In file included from /cygdrive/c/Program Files/HDF_Group/HDF5/1.14.3/include/hdf5.h:21,
                 from /home/christian/old.matio/src/matio_private.h:76,
                 from /home/christian/old.matio/src/endian.c:31:
/cygdrive/c/Program Files/HDF_Group/HDF5/1.14.3/include/H5public.h:285:19: error: conflicting types for ‘ssize_t’; have ‘long long int’
  285 | typedef long long ssize_t;
      |                   ^~~~~~~
In file included from /home/christian/old.matio/src/matio.h:37,
                 from /home/christian/old.matio/src/matio_private.h:36,
                 from /home/christian/old.matio/src/endian.c:31:
/usr/include/stdio.h:81:18: note: previous declaration of ‘ssize_t’ with type ‘ssize_t’ {aka ‘long int’}
   81 | typedef _ssize_t ssize_t;
      |                  ^~~~~~~
```

However, when I install the hdf5-devel Cygwin package, I have the following behavior:

 - If I just call CMake without additional options, it still finds the HDF5 version installed in `C:\Program Files`, not the HDF5 version that is part of Cygwin.
 - If I specify `-DHDF5_ROOT=/usr` when invoking CMake (build directory has to be **clean**, if there's already a CMake configuration there, this won't work), then everything compiles successfully on Cygwin, both with and without my patch, and `./matdump ../share/test_file_hdf5.mat` appears to work as expected.

Since my commit only touches the build system, but not the source files, could it be that the object files from your previous build were still seen as current (because when using `make` instead of e.g. `ninja`, compiler flag changes, such as different include paths, aren't detected, and don't cause a rebuild), which is why you **don't** see the ``ssize_t`` error in your case (because it only tried to re-link, not re-compile). And that because you now installed the official HDF5 installer on your system (but previously hadn't) and CMake detects the Visual Studio one over the Cygwin on, that this broke the Cygwin build, and not my change? And that because you didn't start with a fresh build directory, you didn't see the issues? (Only guessing here.)

tbeu:
OK, after reinstalling libhdf5-devel in CygWin, unsinstalling the MSVC static libs and setting `-DMATIO_SHARED=ON` I can successfully link in CygWin. (It would also work if -lc is set, but not necessary.) CygWin does not install the static hdf5 libs, which is the reason why configuration with `-DMATIO_SHARED=OFF` fails:  Could NOT find HDF5 (missing: HDF5_LIBRARIES) (found version "1.12.3")

Finally, we can leave this as is. Thanks again for fixing and investigating.

Issue #241, opened 2024-03-04
`READ_DATA` macro needs checks before float to int casts


seanm:
Hi @tbeu,

I've been fuzzing lately, and found something that's better to discuss before I attempt any PR. The macro `READ_DATA` has this line:

`data[i + j] = (T)SwapFunc(&v[j]);`

When `v` is float/double it could be outside the range of the integer type `T`.  Like for huge or non-finite floats.  UBSan will then complain of undefined behaviour casting such things to integers.

Probably some range checks will be needed, but that may be tricky as this macro works for many different types...

tbeu:
One idea would be to additionally pass min and max rage values to macros READ_DATA and READ_COMPRESSED_DATA and check if in range before casting.

PR #242, opened 2024-03-04
WIP: issue #241: range checking

seanm:


seanm:
@tbeu something like this?

tbeu:
Yes, with setting `err_` and breaking the loop in the else branch.

tbeu:
Please let mealso know if UBSan is happy with this kind of fixes.

src/read_data.c
```diff
@@ -34,7 +34,9 @@
 #if HAVE_ZLIB
 #include <zlib.h>
 #endif
+#include <float.h>
 #include <stddef.h>
+#include <stdint.h>
```
tbeu, line 39:
```suggestion
#include <limits.h>
```

This fails if `HAVE_STDINT_H` is not defined. Did you mean to include limits.h instead?

seanm: Oh. Doesn't matio require C99? stdint.h I believe is the correct header for `UINT32_MAX` and friends: https://en.cppreference.com/w/c/types/integer

tbeu: No, it's not C99. Go for https://en.cppreference.com/w/c/types/limits instead.

seanm: >No, it's not C99.

Oh. So I guess that's why you have `mat_int32_t` instead of just using `int32_t`?

>Go for https://en.cppreference.com/w/c/types/limits instead.

`UINT32_MAX` is not there.

So I guess we need a custom `MAT_UINT32_MAX` somewhere...

seanm: Or I just use literals like `4294967295U`?

tbeu: You can see the safe-math.h where literals (in hex) are used.

src/read_data.c
```diff
@@ -34,21 +34,30 @@
...
             readcount = fread(v, data_size, len, (FILE *)mat->fp);    \
             if ( readcount == len ) {                                 \
                 for ( i = 0; i < len; i++ ) {                         \
-                    data[i] = (T)v[i];                                \
+                    TT val_ = v[i];                                   \
+                    if (val_ >= min_ && val_ <= max_) {               \
+                        data[i] = (T)val_;                            \
+                    } else {                                          \
+                        break;                                        \
+                    }                                                 \
```
seanm: So I've just realized a problem... when this cast on line 57 is doing float -> int, let's say `uint8_t` specifically, the valid range of floats that can be legally cast to `uint8_t` are [-0.5, 255.5] (I forget if inclusive or exclusive), but `READ_TYPE_MIN` is 0 and `READ_TYPE_MAX` is 255, so we would in fact start rejecting some valid cases.

Not sure best way to do this...

tbeu: For integer conversion you can read http://tzimmermann.org/2018/04/20/safe-integer-conversion-in-c/.
For floating-point to integer conversion we need to deal with rounding. There is https://en.cppreference.com/w/c/numeric/math/round in C99, which rounds to nearest integer.

seanm: @tbeu want me to just take my best stab at it?

tbeu: Sure. That would be appreciated.

seanm: I gave it a try, but have just realised what I've committed is rather wrong... will continue tomorrow.

seanm:
@tbeu could you reopen this?  I have an updated patch...

PR #296, opened 2026-04-08
issue #241: add infrastructure for range checking

### LINKED ISSUE #241: `READ_DATA` macro needs checks before float to int casts

seanm:


seanm:
@tbeu I still need to fill out all the variants of the conversion functions, but any thoughts on this general approach?

seanm:
> @tbeu I still need to fill out all the variants of the conversion functions, but any thoughts on this general approach?

@tbeu friendly ping

tbeu:
CI is failing. This needs to be addressed.

seanm:
> CI is failing. This needs to be addressed.

That's expected. This is WIP. I'd like to know if you find the general approach acceptable before I spend more time to finish it off.

tbeu:
So well, if it works and solves the issue...

seanm:
> So well, if it works and solves the issue...

OK I'll finish it off then...

seanm:
@tbeu ok I've decided to split this into 2 PRs since there's 1 test failing that needs investigation.

So for now, this first PR only adds the the infrastructure to eventually fix the UB conversions. It should change no behaviour. It's ready for review and merging.

tbeu (changes requested):
Appveyor fails.

seanm:
> Appveyor fails.

Hmmm,

```
  ..\..\src\read_data.c(421): error C2054: expected '(' to follow 'inline' [C:\projects\matio\visual_studio\libmatio\libmatio.vcxproj]
  ..\..\src\read_data.c(422): error C2085: 'ConvertFromUInt16ToUInt16' : not in formal parameter list [C:\projects\matio\visual_studio\libmatio\libmatio.vcxproj]
  ..\..\src\read_data.c(422): error C2143: syntax error : missing ';' before '{' [C:\projects\matio\visual_studio\libmatio\libmatio.vcxproj]
```

Is this an old compiler? Or one that doesn't support C very well?  Maybe it does not like `static` and `inline` together?

seanm:
Some searching suggests indeed MS compilers don't support C well, and might not like `static inline`.  I've just removed the `inline` keyword, we'll see if that fixes it... Such a tiny function will probably be inlined anyway.

src/read_data_impl.h
```diff
@@ -21,6 +21,77 @@
 #define READ_TYPE_UINT64_DATA CAT(READ_TYPED_FUNC1, UInt64)
 #endif /* HAVE_MAT_UINT64_T */
 
+// Converts value from its own type to READ_TYPE, carefully ensuring the value fits, otherwise sets success to false.
```
tbeu:
```suggestion
// Stubs only: Converts value from its own type to READ_TYPE.
```

seanm: OK, done. Indeed that comment is not yet true, but it will be in the next PR.

tbeu:
Superseded by #327.

PR #300, opened 2026-05-06
Add 3 read-path fuzzer harnesses, expanded dict, and CIFuzz

Tomer-PL:
## Summary

- Add three new **read-only** fuzzer harnesses targeting code paths the existing `matio_fuzzer` does not reach:
  - `matio_read_fast_fuzzer` — skips `Mat_VarPrint(v, 1)` to avoid per-input timeouts on large-dims files; exercises `Mat_VarReadDataAll`, `Mat_VarDuplicate`, and `Mat_VarRead`-by-name
  - `matio_hyperslab_fuzzer` — drives `Mat_VarReadData` / `Mat_VarReadDataLinear` with fuzz-derived start/stride/edge (bounded to satisfy API contracts); also exercises `Mat_CalcSingleSubscript` and related helpers
  - `matio_struct_cell_fuzzer` — exercises `Mat_VarGetCells`, `GetCellsLinear`, `GetStructFieldByIndex`, `GetStructFieldByName`, `GetStructs`, and `GetStructsLinear` against attacker-controlled `.mat` files
- Expand `matio_fuzzer.dict` from 4 → 40 entries (HDF5 magic, v5 type/class tags, zlib headers, MCOS markers, v4 type bytes)
- Update `build.sh`: C++17, stale-object cleanup between sanitizer builds, `matio_test_datasets` in seed corpus, dict propagated to all fuzzers
- Extend `.github/workflows/fuzz.yml` with a CIFuzz job (600s fuzzing on each PR via `google/oss-fuzz` CIFuzz actions)

No write-path fuzzers are included — all harnesses target read paths only. The existing `matio_fuzzer` is unchanged.

Total new code: ~530 lines (3 harnesses + build/workflow changes).


.github/workflows/fuzz.yml
```diff
@@ -33,3 +37,26 @@ jobs:
...
+          oss-fuzz-project-name: 'matio'
+          language: c++
+      - name: Run Fuzzers
+        uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+        with:
+          oss-fuzz-project-name: 'matio'
+          language: c++
+          fuzz-seconds: 600
+      - name: Upload Crash
+        uses: actions/upload-artifact@v4
```
tbeu:
```suggestion
        uses: actions/upload-artifact@v7
```

tbeu:
Thanks for the PR.

Never mind the failing [Autotools build / linux-gcc (pull_request)](https://github.com/tbeu/matio/actions/runs/25432140804/job/74661989479?pr=300) that is due to PRs from forks cannot access repo secrets.

Why is [Ossfuzz build / cifuzz (pull_request)](https://github.com/tbeu/matio/actions/runs/25432140822/job/74640663184?pr=300) failing?

I guess I am going to align fuzzers with current fuzzer and make individual fuzzers instead. I was not ware of the timing issue with `Mat_VarPrint`.

Tomer-PL:
Thanks for the review!

The **cifuzz** failure is actually expected and a good sign: `matio_struct_cell_fuzzer` found a reproducible pre-existing bug on its first run (the leak in `Mat_VarGetStructsLinear` that we reported). CIFuzz tried to compare against the previous OSS-Fuzz build, but since `matio_struct_cell_fuzzer` is a new target that doesn't exist upstream yet, it couldn't determine whether the crash is a regression vs. pre-existing — so it conservatively reports failure.

Once the harnesses are merged and the first OSS-Fuzz build includes them, subsequent CIFuzz runs will correctly diff against the baseline and only fail on actual regressions introduced by a PR.

> I guess I am going to align fuzzers with current fuzzer and make individual fuzzers instead.

That's exactly the right approach — multiple targeted fuzzers is the recommended practice per Google's [fuzz target guide](https://github.com/google/fuzzing/blob/master/docs/good-fuzz-target.md): *"Usually, the narrower the target the better. E.g. if your target can parse several data formats, split it into several targets, one per format."* and *"For any large API prefer to create several fuzz targets that cover subsets of the API."*

The reasoning is that fuzzing has super-linear complexity — a single target covering too many code paths dilutes libFuzzer's coverage signal and gets stuck. Our three new fuzzers follow this pattern: each targets a distinct API surface the original doesn't reach (hyperslab/partial reads, struct/cell accessor APIs, and the fast read-without-print path).

> I was not aware of the timing issue with `Mat_VarPrint`.

Yeah, `Mat_VarPrint(var, 1)` iterates over every element, so a malformed file claiming huge dims can wedge the fuzzer for its entire timeout budget before any other inputs get tried. The `matio_read_fast_fuzzer` avoids this by using `Mat_VarPrint(var, 0)` (metadata only).

PR #308, opened 2026-06-17
Added cmake install package

techsupport-noah:
I added this to install matio as a cmake package.
Using find_package on the installed config worked for me when using with a static lib as well as an executable.
For that i also added code to differentiate between build time and install time.

Other than that i linked the hdf5 wrapper target private instead of public, as matio does not build hdf5 itsself and does not depend on it in the public api afaik.

### DIFF
--- a/cmake/src.cmake
+++ b/cmake/src.cmake
@@ -62,11 +62,12 @@ endif()
 add_library(${PROJECT_NAME}::${PROJECT_NAME} ALIAS ${PROJECT_NAME})
 
 target_include_directories(${PROJECT_NAME}
-    INTERFACE ${PROJECT_SOURCE_DIR}/src
-    PUBLIC    ${PROJECT_BINARY_DIR}/src
+    INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>
+    PUBLIC    $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/src>
+              $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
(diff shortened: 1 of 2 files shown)


MaartenBent (commented):
Nice addition!

However It does not work for me (static matio with static hdf5 libs), because the installed package depends on the internal `MATIO::HDF5` and `MATIO::ZLIB` wrapper libraries. I can fix that so it uses the actual libraries instead.

I have some formatting remarks as well, like using 4 spaces and end the file with a new-line. And move `matio.cmake.in` into the cmake folder.

The last problem is that it breaks building `test_mat` (see all the CI errors). Because this Implicitly includes hdf5 and zlib through matio. After making hdf5 private, it does not inherit the include directories and libraries anymore.

I've created a PR on top of this one with my suggested changes, see #309

techsupport-noah:
> Nice addition!
> 
> However It does not work for me (static matio with static hdf5 libs), because the installed package depends on the internal `MATIO::HDF5` and `MATIO::ZLIB` wrapper libraries. I can fix that so it uses the actual libraries instead.
> 
> I have some formatting remarks as well, like using 4 spaces and end the file with a new-line. And move `matio.cmake.in` into the cmake folder.
> 
> The last problem is that it breaks building `test_mat` (see all the CI errors). Because this Implicitly includes hdf5 and zlib through matio. After making hdf5 private, it does not inherit the include directories and libraries anymore.
> 
> I've created a PR on top of this one with my suggested changes, see #309

Thanks for changing it in a way that it will work for everyone!
As i bring my own hdf5 lib when compiling matio, i did not notice the problems regarding the missing hdf5 lib.

PR #309, opened 2026-06-17
Improve CMake install package

MaartenBent:
Some fixes on top of #308

And added a function to set warning related properties for each target.


MaartenBent:
I'll fix the pre-commit lint error later.

The other errors happen because conan (i think) wants to use `Visual Studio 18 2026` which CMake cannot find.

tbeu:
@techsupport-noah Could you test this PR? OK for you to merge this one in favour of #308?

techsupport-noah:
@tbeu Works for me! I will close the other MR. Thanks for improving @MaartenBent!

PR #310, opened 2026-06-18
OSS-Fuzz: Add new corpora seeds for fuzzers

arthurscchan:
This PR add more corpora seeds for fuzzers to targets different version of mat.


tbeu:
Why another fuzzer? Aren't the seeds just good enough?

arthurscchan:
Sorry for the late reply. Indeed, I am overthinking it. I have done some local test and seed/corpora does help in exploring those logic and unblock the coverage. I will rebase this PR and adds in the necessary seed/corpora. Sorry again for the rush conclusion.

tbeu:
FYI: Not a single new OSS fuzzer issue was raised with the new seeds within the last two weeks after merge. I guess it already was covered pretty well.

PR #311, opened 2026-06-21
Enable support for zlib-ng in compat mode

tbeu:
@MaartenBent Kindly asking for review, espcially the FetchContent fallback which is currently not found for other dependencies. Thank you.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -172,6 +172,29 @@ jobs:
             cmake --build build -- -j8
             ctest --no-tests=error --test-dir build --build-config Release --parallel 8
 
+  build-zlib-ng:
+    name: linux-gcc (zlib-ng)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 5
+          submodules: recursive
+      - name: Setup python environment
+        uses: actions/setup-python@v6
+        with:
+          python-version: 3.13
+      - name: Install python packages
+        run: pip install --disable-pip-version-check --user "conan>=2.16.1,<3"
+      - name: Configure with zlib-ng (zlib compat mode)
+        run: |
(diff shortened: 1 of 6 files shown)


MaartenBent (commented):
Looks good overall. I tried it on Windows/msvc with fetching zlib-ng. I haven't tried with system libraries.

- When zlib-ng is configured, I get a warning `ZLIB_ENABLE_TESTS is deprecated. Please use BUILD_TESTING instead`. It would be better to use `EXCLUDE_FROM_ALL` so only the used targets from zlib-ng will be built. `EXCLUDE_FROM_ALL` also excludes the targets from being installed, so we need to do that ourselves.
- It is not necessary to use cached variables.
- I think it should build and use the static zlib-ng libraries when using static matio and shared zlib-ng libraries when using shared matio, like is done for hdf5 and legacy zlib.
- Since we're building zlib-ng ourselves, we can  use their target names instead of the zlib aliases. This can prevent problems when CMake already knows about legacy zlib targets.
- Use the release tag instead of the commit hash so it is clearer which version is used.

All together, this would look like:
```cmake
        # Fallback: fetch and build zlib-ng from source (always in compat mode)
        message(STATUS "zlib-ng not found on system, fetching from GitHub")
        FetchContent_Declare(zlib-ng
            GIT_REPOSITORY "https://github.com/zlib-ng/zlib-ng"
            GIT_TAG        "2.3.3"
            GIT_SHALLOW    TRUE
            EXCLUDE_FROM_ALL
        )
        set(ZLIB_COMPAT ON)
        FetchContent_MakeAvailable(zlib-ng)

        if(NOT MATIO_SHARED AND TARGET zlib-ng-static)
            MATIO_CREATE_ZLIB(zlib-ng-static)
            install(TARGETS zlib-ng-static)
        elseif(MATIO_SHARED AND TARGET zlib-ng)
            MATIO_CREATE_ZLIB(zlib-ng)
            install(TARGETS zlib-ng)
        else()
            message(FATAL_ERROR "Failed to build zlib-ng in compat mode")
        endif()
```

Maybe the first case (when zlib-ng is found on the system) should also depend on `MATIO_SHARED`, and prefer static libraries in this case.

tbeu:
@MaartenBent Thanks for the review. Proposed changes are applied.

PR #312, opened 2026-06-25
bound sparse ir/jc/data reads to allocation in Mat_VarRead73

naruto-lgtm:
Repro: open a v7.3 file whose sparse `ir`, `jc`, or `data` subdataset is stored with rank > 1, so its element count exceeds `dims[0]`.
Cause: the `MAT_C_SPARSE` path in `Mat_VarRead73` sizes each buffer from `dims[0]` but `H5Dread` fills the whole dataset (product of all dims), overflowing the heap allocation.
Fix: require each subdataset to be a 1-D vector whose length matches the allocation before reading it.

Confirmed with ASAN: a 2-D `ir` turns an 8-byte allocation into a 24-byte write before the patch, and is rejected cleanly after, with the sparse round-trip tests still passing.

### DIFF
--- a/src/mat73.c
+++ b/src/mat73.c
@@ -3203,6 +3203,13 @@ Mat_VarRead73(mat_t *mat, matvar_t *matvar)
                     size_t nbytes = 0;
                     sparse_data->nir = (mat_uint32_t)dims[0];
                     free(dims);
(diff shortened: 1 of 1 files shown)


tbeu:
Thanks for the PR. Applied as is.

PR #313, opened 2026-07-01
Bound v4 sparse data buffer to the read element size in Mat_VarRead4

naruto-lgtm:
Repro: read a v4 file whose sparse variable declares a non-double data type (e.g. uint8) when the library is built without extended sparse (`--disable-extended-sparse` / `-DMATIO_EXTENDED_SPARSE=OFF`). ASAN reports a heap-buffer-overflow WRITE of size 8 in `ReadDoubleData` off the `malloc` at mat4.c:643, via `Mat_VarReadDataAll` -> `Mat_VarRead4`.

Cause: the `MAT_C_SPARSE` branch sizes `sparse->data` as `ndata * Mat_SizeOf(data_type)` using the type declared in the file, but in a non-extended build the values are always read through `ReadDoubleData`, which writes `ndata` doubles regardless of that type. A declared uint8 buffer is `ndata` bytes while `ndata*8` bytes get written.

Fix: force the destination type to double before sizing in the non-extended build, matching what `Mat_VarRead5` already does, and route both the complex and non-complex sparse allocations through the overflow-checked `Mul()`.

### DIFF
--- a/src/mat4.c
+++ b/src/mat4.c
@@ -496,9 +496,24 @@ Mat_VarRead4(mat_t *mat, matvar_t *matvar)
                 }
                 sparse->ndata = sparse->nir;
                 data_type = matvar->data_type;
+#if !defined(EXTENDED_SPARSE)
+                /* Without extended sparse support the values are always read as
+                   double, so the destination buffer must be sized for double
+                   regardless of the type declared in the file. */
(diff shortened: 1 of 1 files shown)


tbeu:
Thanks a lot @naruto-lgtm

PR #314, opened 2026-07-02
bound GetDataSlab whole-array memcpy to a full column selection

naruto-lgtm:
Repro: read a partial column range (edge[1] < dims[1]) from a compressed 2-D numeric struct field or cell element with Mat_VarReadData.
Cause: the rank-2 fast path in GetDataSlab copies the whole variable (matvar->nbytes) but only guards edge[0] == dims[0], so a caller buffer sized to the selection (edge[0]*edge[1]) is overrun when fewer columns are requested.
Fix: take the whole-buffer memcpy only when the second dimension is fully selected too; otherwise fall through to the element-wise slab copy.

ASAN on the pre-read (in-memory) path shows a write of the full dims[0]*dims[1] element count into the edge-sized output at mat5.c:4784 via Mat_VarReadData5 -> GetDataSlab; with the guard the requested columns read back correctly and the run stays clean.


tbeu:
Thank you again! Appreciated.

PR #315, opened 2026-07-03
Validate read range up front in Mat_VarReadDataLinear5

naruto-lgtm:
Repro: open a compressed file whose struct field or cell element is small enough to be pre-read into internal->data, then call Mat_VarReadDataLinear with start/edge past the element count.
Cause: Mat_VarReadDataLinear5 range-checks stride/edge/start only after it dispatches the already-read fast path, so GetDataLinear copies past the pre-read buffer.
Fix: hoist the check ahead of the fast path, the way Mat_VarReadDataLinear4 already does.


tbeu:
Appreciated. Thank you.

PR #316, opened 2026-07-04
validate edge and stride in fixed-rank slab readers

naruto-lgtm:
Repro: Mat_VarReadData on an uncompressed v4 rank-2 double array with edge[1] negative.
Cause: ReadDataSlab1/ReadDataSlab2 (and their compressed twins) never run the CheckEdgeOverflow/CheckSlabOverflow checks that ReadDataSlabN does, so the ReadDataSlab2 read-all fast path turns (ptrdiff_t)edge[0]*edge[1] into a huge size_t length and the direct fread fills the smaller caller buffer from the larger file data (ASan heap-buffer-overflow write in ReadDoubleData). Mat_VarReadData4 sets its rank-2 range-check error but still calls ReadDataSlab2, so the callee is where the invariant belongs.
Fix: run the same edge/slab overflow checks (and reject negative start/stride/edge in the rank-1 readers) up front in all four fixed-rank slab readers, matching ReadDataSlabN. Valid slab reads are unchanged.

### DIFF
--- a/src/read_data.c
+++ b/src/read_data.c
@@ -909,6 +909,10 @@ ReadDataSlab1(mat_t *mat, void *data, enum matio_classes class_type, enum matio_
     int err, i;
     size_t data_size;
 
+    if ( start < 0 || stride < 0 || edge < 0 ) {
+        return MATIO_E_BAD_ARGUMENT;
+    }
+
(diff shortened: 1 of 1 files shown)


tbeu:
Thx!

PR #318, opened 2026-07-11
validate edge product in Mat_VarGetCells and Mat_VarGetStructs

naruto-lgtm:
Repro: Mat_VarGetCells on a rank-2 cell array with edge={3, 1431655766}.
Cause: N, the product of edge[], is accumulated in a signed int with no overflow guard. 3*1431655766 wraps to 2, so the pointer array is malloc'd for two cells while the copy loop still writes cells[0..2], one past the allocation (ASan heap-buffer-overflow in matvar_cell.c). Mat_VarGetStructs sizes its field array from the same unchecked product.
Fix: reject a negative edge or a product past INT_MAX before the allocation in both, matching CheckEdgeOverflow in the fixed-rank slab readers. Valid selections are unchanged.

### DIFF
--- a/src/mat.c
+++ b/src/mat.c
@@ -19,6 +19,7 @@
 #endif
 #include "safe-math.h"
 #include <assert.h>
+#include <limits.h>
 #include <stdlib.h>
 #include <string.h>
 #include <stdio.h>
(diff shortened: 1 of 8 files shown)


naruto-lgtm:
any update?

PR #324, opened 2026-07-17
bound MATLAB_class and MATLAB_sparse attribute reads to one element

naruto-lgtm:
Repro: open a v7.3 file whose MATLAB_class attribute (read for every variable) or MATLAB_sparse attribute (on a struct/sparse group) is stored with a simple dataspace of two or more elements.
Cause: both are read with a bare H5Aread into a single-element destination, calloc(class_len+1) for the class string in Mat_H5ReadVarInfo and a stack `unsigned nrows` in Mat_H5ReadGroupInfo, but H5Aread copies the whole attribute, so a multi-element attribute writes past the destination during Mat_VarReadNextInfo. ASan shows a heap write of 14 bytes at mat73.c:591 and a stack write of 8 bytes at mat73.c:1068.
Fix: read both through Mat_H5ReadScalarAttribute, which already rejects a non-scalar attribute before the copy, the same guard the MATLAB_global, MATLAB_int_decode and MATLAB_empty reads use.

### DIFF
--- a/src/mat73.c
+++ b/src/mat73.c
@@ -588,13 +588,13 @@ Mat_H5ReadVarInfo(matvar_t *matvar, hid_t dset_id)
     }
     class_str = (char *)calloc(class_len + 1, 1);
     if ( NULL != class_str ) {
-        herr_t herr = H5Aread(attr_id, mem_type_id, class_str);
-        if ( herr < 0 ) {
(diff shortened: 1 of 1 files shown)


naruto-lgtm:
any update?

tbeu:
Seen it.

PR #327, opened 2026-08-10
saturate float-to-integer conversion in the typed data readers

### LINKED ISSUE #241: `READ_DATA` macro needs checks before float to int casts

naruto-lgtm:
Repro: read any v4 sparse matrix whose row/column index (stored as double) falls outside uint32 range; ReadUInt32Data casts it straight into mat_uint32_t.
Cause: the typed readers in read_data_impl.h convert a double/single element to an integer destination with a plain (T)v[i] cast, which is undefined for an out-of-range or NaN value (C11 6.3.1.4). Reachable from Mat_VarRead4/Mat_VarRead5 for any floating-stored, integer-typed array.
Fix: saturate a floating source to the destination range (NaN maps to 0) before the narrowing cast, in the uncompressed and compressed readers alike; integer sources keep the existing modular cast, so every in-range value stays byte-identical (full ctest suite still passes).

Fixes #241

### DIFF
--- a/src/read_data.c
+++ b/src/read_data.c
@@ -23,7 +23,25 @@
 #include <math.h>
 #include <time.h>
 
-#define READ_DATA_NOSWAP(T)                                           \
+/* Conversion of a value read from the file into the destination type T.
+ * Casting an out-of-range (or NaN/Inf) floating value to an integer type is
+ * undefined behavior (C11 6.3.1.4), which a crafted file can reach whenever a
+ * floating source type is read into an integer destination (e.g. v4 sparse
+ * row/column indices are stored as double and read into mat_uint32_t). Integer
+ * sources keep the existing modular cast; floating sources saturate to the
+ * destination range and map NaN to 0, leaving every in-range value unchanged. */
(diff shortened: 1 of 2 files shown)


tbeu:
Does this fix #241?

naruto-lgtm:
Yes, #241 is exactly this cast. The double/single instantiations of all four macro variants (READ_DATA, READ_DATA_NOSWAP and their compressed twins) now clamp to the destination range before converting, with NaN going to 0, and the slab readers are covered too since they go through the same typed functions. It follows the range-check idea you suggested on the issue, just passed in as a conversion macro so integer sources keep the plain cast. I've added Fixes #241 to the description so it closes on merge.

tbeu:
Thanks a lot! Appreciated!

PR #329, opened 2026-08-17
Find CMake zlib target of hdf5 2.2.0

MaartenBent:
The CMake target for zlib in `hdf5-2.2.0-win-vs2026_cl` is called `ZLIB::ZLIBSTATIC`.


tbeu:
Thanks.

@Vollstrecker Anything else to consider for latest zlib?

Vollstrecker:
I don't know conan, from target side this should be enough.

PR #332, opened 2026-08-21
Fix rank-N in-memory slab indexing

### LINKED ISSUE #331: GetDataSlab: rank>2 slab reads return the wrong elements and read out of bounds
Two independent defects in the rank>2 slab path (`GetDataSlab` in `src/mat5.c`). Both are reachable from an ordinary `Mat_VarReadData` call whose selection is in range on every dimension, and both are present on master (`d410ad4`) as well as 1.5.30. Depending on the dimensions, the result is either silently wrong data or an out-of-bounds read.

The path only applies to a variable whose data is already in memory -- `matvar->compression == MAT_COMPRESSION_ZLIB` and `matvar->internal->data != NULL`, i.e. a small field inside a compressed struct or cell that `ReadNextStructField` / `ReadNextCell` inflated during `Mat_VarReadNextInfo`. A top-level uncompressed variable goes through `ReadDataSlabN` instead.

## 1. `dimp[]` is not the cumulative dimension product

In `GET_DATA_SLABN` (`src/mat5.c`):

```c
dimp[i] = (int)dims[i - 1];
for ( j = i; j--; ) {
    inc[i] *= (int)dims[j];
    dimp[i] *= (int)dims[j + 1];
}
...
I += dimp[i - 1] * start[i];
```

The inner loop multiplies `dims[i], dims[i-1], ..., dims[1]` onto an initial `dims[i-1]`, so `dimp[i]` ends up as `dims[i-1] * (dims[1]*...*dims[i])` rather than the cumulative product `dims[0]*...*dims[i]` the base offset needs -- too large by a factor of `dims[i-1]/dims[0]`. Whenever `dims[0] < dims[i-1]` and `start[i] > 0`, `I` points at the wrong element, and often past the end of the buffer.

`stride` plays no part in this: the same faults occur with `stride` all 1.

### 1a. Silent wrong data (no sanitizer needed)

```c
#include <stdio.h>
#include <stdlib.h>
#include <matio.h>

int main(void) {
    size_t dims[4] = {2, 1, 1, 2};
    double values[4] = {0, 1, 2, 3};
    remove("wrong_element.mat");
    mat_t *mat = Mat_CreateVer("wrong_element.mat", NULL, MAT_FT_MAT5);
    matvar_t *leaf = Mat_VarCreate("f0", MAT_C_DOUBLE, MAT_T_DOUBLE, 4, dims, values, 0);
    matvar_t *leaves[2] = {leaf, NULL};
    size_t sdims[2] = {1, 1};
    matvar_t *top = Mat_VarCreate("s", MAT_C_STRUCT, MAT_T_STRUCT, 2, sdims, leaves, 0);
    Mat_VarWrite(mat, top, MAT_COMPRESSION_ZLIB);
    Mat_VarFree(top);
    Mat_Close(mat);

    mat = Mat_Open("wrong_element.mat", MAT_ACC_RDONLY);
    matvar_t *readback = Mat_VarReadNextInfo(mat);
    matvar_t *field = Mat_VarGetStructFieldByName(readback, "f0", 0);

    int start[4] = {0, 0, 0, 1}, stride[4] = {1, 1, 1, 1}, edge[4] = {1, 1, 1, 1};
    double out = -1.0;
    int err = Mat_VarReadData(mat, field, &out, start, stride, edge);
    printf("err=%d, element (0,0,0,1) of 2x1x1x2: want 2.0, got %.1f\n", err, out);

    Mat_VarFree(readback);
    Mat_Close(mat);
    return 0;
}
```

Expected `want 2.0, got 2.0`. Actual, on 1.5.30 and on master:

```
err=0, element (0,0,0,1) of 2x1x1x2: want 2.0, got 1.0
```

`dimp[2]` is 1 where the cumulative product is 2, so the read lands on element 1 instead of element 2. `Mat_VarReadData` reports success.

### 1b. Out-of-bounds read

Same defect, dimensions chosen so the wrong offset leaves the buffer: a rank-5 `MAT_C_SINGLE` field of dims `{1,1,4,2,4}` (32 elements, 128 bytes), asking for the single element at `(0,0,0,0,2)`. `dimp[3]` is 32 where the cumulative product is 8, so the base offset is element 64 of 32.

With matio built `-fsanitize=address`:

```
==29==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60c000000440
READ of size 4 at 0x60c000000440 thread T0
    #0 in GetDataSlab src/mat5.c:4878
    #1 in Mat_VarReadData5 src/mat5.c
    #2 in main repro.c:65
0x60c000000440 is located 128 bytes to the right of 128-byte region
allocated by thread T0 here:
    #0 in malloc
    #1 in Mat_VarRead5 src/mat5.c:4514
    #2 in ReadNextStructField src/mat5.c:1772
```

(Line numbers from master `d410ad4`; on 1.5.30 the same report points at `mat5.c:4525` and `mat5.c:4163`.) The overrun buffer is the one matio allocated for the field itself, correctly sized at `prod(dims) * sizeof(float)`.

## 2. The rank-loop wrap correction over-advances on a partial selection

In `GET_DATA_SLABN_RANK_LOOP`:

```c
if ( (I % dimp[j]) != 0 ) {
    ptr_in += dimp[j] - (I % dimp[j]) + dimp[j - 1] * start[j];
    I      += dimp[j] - (I % dimp[j]) + dimp[j - 1] * start[j];
}
```

This assumes a just-finished dimension-*j* block ends on a multiple of `dimp[j]`, which only holds when every inner dimension is traversed in full.
With a partial inner selection (`edge[k] < dims[k]`, or `start[k] != 0`) the correction over-advances and the walk leaves the buffer. This is independent of defect 1 -- it still happens with a corrected `dimp[]`.

A rank-4 `MAT_C_UINT8` field of dims `{4,4,2,4}` (128 elements), with `start {2,3,0,1}`, `stride {1,2,1,1}`, `edge {1,1,2,3}` -- in range on every
dimension, `{2,3,1,3} <= {3,3,1,3}` -- should read source indices `46 62 78 94 110 126`. It walks `46 62 98 126 162 190` instead, and 128 is the
end of the array:

```
==32==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60c0000003e2
READ of size 1 at 0x60c0000003e2 thread T0
    #0 in GetDataSlab src/mat5.c:4922
0x60c0000003e2 is located 34 bytes to the right of 128-byte region
```

## Notes

- Every selection above satisfies `stride[i]*(edge[i]-1)+start[i] <= dims[i]-1`   for all `i`, and the output buffer is `prod(edge)` elements of
  `Mat_SizeOfClass(class_type)`, which is how `tools/matdump.c` sizes it.
- The equivalent expressions appear in `src/read_data.c` (`ReadDataSlabN` and   its compressed variant) as well. I have not tested those paths; there the   effect would presumably be wrong file offsets rather than a heap overrun.
- Found while fuzzing matio through the public API, then reduced to the   programs above; both build against an unmodified matio.

carrerasdarren-cell:
## Summary

- validate each rank-N selection before indexing pre-read compressed data
- replace the incremental wrap correction with direct column-major block indexing
- calculate dimension and selection products in `size_t` with checked multiplication

Fixes #331.

## Root cause

`GetDataSlab` did not build `dimp[]` as cumulative dimension products, so higher-dimensional starts could select the wrong element or move beyond the input buffer. Its modulo-based wrap correction also assumed each inner dimension was selected in full, causing partial selections to over-advance.

The replacement computes each selected inner block from its explicit rank coordinates, then walks dimension zero with the requested stride. This avoids carrying pointer state between blocks.

## Testing

Built with AddressSanitizer and UndefinedBehaviorSanitizer and replayed the public issue shapes through `Mat_VarReadData` on compressed struct fields:

- rank-4 wrong-element case: expected and returned `2`
- rank-5 base-offset case: expected and returned `16`, without the prior 128-byte out-of-bounds read
- rank-4 partial-selection case: returned `46, 62, 78, 94, 110, 126`, without the prior 34-byte out-of-bounds read
- deterministic differential control: 10,000 valid rank-5 start/stride/edge selections matched an independent reference indexer
- full ASan/UBSan CTest matrix: 6,256/6,256 passed
- `clang-format --dry-run --Werror src/mat5.c` and `git diff --check` passed

### DIFF
--- a/src/mat5.c
+++ b/src/mat5.c
@@ -4533,27 +4533,6 @@ Mat_VarRead5(mat_t *mat, matvar_t *matvar)
 }
 
 #if HAVE_ZLIB
-#define GET_DATA_SLABN_RANK_LOOP                                                \
-    do {                                                                        \
-        for ( j = 1; j < rank; j++ ) {                                          \
-            cnt[j]++;                                                           \
-            if ( (cnt[j] % edge[j]) == 0 ) {                                    \
-                cnt[j] = 0;                                                     \
-                if ( (I % dimp[j]) != 0 ) {                                     \
-                    ptr_in += dimp[j] - (I % dimp[j]) + dimp[j - 1] * start[j]; \
-                    I += dimp[j] - (I % dimp[j]) + dimp[j - 1] * start[j];      \
-                } else if ( start[j] ) {                                        \
-                    ptr_in += dimp[j - 1] * start[j];                           \
-                    I += dimp[j - 1] * start[j];                                \
-                }                                                               \
-            } else {                                                            \
-                I += inc[j];                                                    \
-                ptr_in += inc[j];                                               \
-                break;                                                          \
-            }                                                                   \
-        }                                                                       \
-    } while ( 0 )
-
 #define GET_DATA_SLAB2(T)                              \
     do {                                               \
         ptr_in += start[1] * dims[0] + start[0];       \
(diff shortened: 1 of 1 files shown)


tbeu:
Thanks for fixing though I wonder what is your motivation since that feature seems to be hardly used.

carrerasdarren-cell:
Thanks for asking. I picked this up from the public OSS-Fuzz issue #331. Although the in-memory rank-N slab path appears uncommon, the reproducer exposed both silent wrong-element output and heap over-reads. I reproduced those locally and differentially checked 10,000 selections against the file-backed path, so the motivation was to make the existing API behavior correct and memory-safe rather than to add a new feature.

PR #336, opened 2026-08-29
Size opaque metadata buffer by checked element count in Mat_VarRead73

naruto-lgtm:
Repro: read a v7.3 file with an unknown-class integer dataset (classified MAT_C_OPAQUE) whose dataspace element count is near the integer limits.
Cause: the opaque branch sizes the uint32 metadata buffer with a raw `(size_t)nelems * sizeof(mat_uint32_t)`, but H5Dread fills `nelems` elements from the H5S_ALL selection, so the byte multiply wraps (and `nelems` truncates on a 32-bit size_t) and H5Dread writes past the allocation.
Fix: size the buffer with the overflow-checked `Mul` used by the other readers and skip the read when the count would overflow or truncate, keeping the allocation in step with the `H5Screate_simple` element count.

### DIFF
--- a/src/mat73.c
+++ b/src/mat73.c
@@ -3569,9 +3569,14 @@ Mat_VarRead73(mat_t *mat, matvar_t *matvar)
                         H5Sclose(space_id);
                     }
                     if ( nelems > 0 ) {
-                        /* Read the raw data as uint32 values */
-                        mat_uint32_t *meta =
-                            (mat_uint32_t *)malloc((size_t)nelems * sizeof(mat_uint32_t));
+                        /* Size the buffer with a checked multiply so it matches
+                         * the element count passed to H5Screate_simple below. */
+                        size_t meta_size = 0;
+                        mat_uint32_t *meta = NULL;
+                        if ( (hsize_t)(size_t)nelems == nelems &&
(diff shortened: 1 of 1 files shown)


naruto-lgtm:
any update?

src/mat73.c
```diff
@@ -3569,9 +3569,14 @@ Mat_VarRead73(mat_t *mat, matvar_t *matvar)
...
                     if ( nelems > 0 ) {
-                        /* Read the raw data as uint32 values */
-                        mat_uint32_t *meta =
-                            (mat_uint32_t *)malloc((size_t)nelems * sizeof(mat_uint32_t));
+                        /* Size the buffer with a checked multiply so it matches
+                         * the element count passed to H5Screate_simple below. */
+                        size_t meta_size = 0;
+                        mat_uint32_t *meta = NULL;
+                        if ( (hsize_t)(size_t)nelems == nelems &&
+                             0 == Mul(&meta_size, (size_t)nelems, sizeof(mat_uint32_t)) ) {
```
tbeu:
```suggestion
                             MATIO_E_NO_ERROR == Mul(&meta_size, (size_t)nelems, sizeof(mat_uint32_t)) ) {
```

naruto-lgtm: Done, switched the comparison to MATIO_E_NO_ERROR to match the other Mul call sites.

Issue #350, opened 2026-09-13
stack-overflow: unbounded recursion in ReadNextCell


kingkingkingson:
## Summary

`matdump -d` on a crafted MAT5 cell nest recurses `ReadNextCell` → `ReadNextCell` at `mat5.c:1495` until the stack is exhausted. ASan reports `stack-overflow`. There is no depth / nesting limit on cell-in-cell parsing.

## Affected version

- Repository: https://github.com/tbeu/matio
- Version: **1.6.0** (cloned `master` HEAD)
- Branch: `master`
- Commit: `b1355a2b4532766d04eafa00299168c76fcf452f`
- CWE: CWE-674 / CWE-400
- Error class: **stack-overflow (unbounded recursion)**
- Site: `ReadNextCell / src/mat5.c:1495 (self-recursion)`

## Build

Need zlib for compressed MAT5. AddressSanitizer makes the stack overflow obvious. This hit was offtarget for a different hypothesis (`ReadOpaqueInfo5` integer wrap) but the sanitizer report is a real, unique crash in `ReadNextCell`.

```bash
sudo apt-get install -y cmake gcc libhdf5-dev zlib1g-dev
git clone https://github.com/tbeu/matio.git && cd matio
git checkout b1355a2b4532766d04eafa00299168c76fcf452f

cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo \
  -DCMAKE_INSTALL_PREFIX="$PWD/install" \
  -DCMAKE_C_FLAGS="-fsanitize=address,undefined -fno-omit-frame-pointer -g" \
  -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address,undefined" \
  -DCMAKE_SHARED_LINKER_FLAGS="-fsanitize=address,undefined"
cmake --build build --parallel
cmake --install build
```

## Reproduce

```bash
export PREFIX="$PWD/install"
export ASAN_OPTIONS="detect_leaks=1:halt_on_error=0:allocator_may_return_null=1"

"$PREFIX/bin/matdump" -d input.mat
```

Expected: `SUMMARY: AddressSanitizer: stack-overflow` with a long `ReadNextCell` self-recursion chain.

## Sanitizer evidence

```
==1001==ERROR: AddressSanitizer: stack-overflow on address 0x7ffd09570ff8 (pc 0x7f078d8b1da6 bp 0x7ffd09571860 sp 0x7ffd09571000 T0)
[...]

PR #353, opened 2026-09-13
Limit cell/struct nesting depth in ReadNextCell

kingkingkingson:
As reported by https://github.com/tbeu/matio/issues/350


tbeu (changes requested):
The implementation uses a file-scope mutable global static int nest_depth. This is:

 - Not thread-safe — two files read in parallel corrupt the shared count (premature limit triggers or under-counted nesting).
 - Not re-entrant if parsing is re-entered.

The codebase already has the correct idiom for this exact problem: mat->ref_depth in struct _mat_t (matio_private.h:122), which stores per-file recursion depth for the analogous MAT73/HDF5 reference recursion. Since the mat pointer is threaded through every one of these functions, the PR should store the depth in the mat struct (e.g. a new int nest_depth field in struct _mat_t, or reuse the existing ref_depth field) instead of a module global. That avoids both races and surprises, and matches existing style.

PR #364, opened 2026-09-16
Fix MCOS enum decoding

### LINKED ISSUE #363: Reading Simulink.IntEnumType data objects
Hi all,

For Simulink enumeration data objects (Simulink.IntEnumType), MATLAB appears to be using a special format when storing them in a .mat file.

### Version

This was reproduced on Windows using a Visual Studio build (with MCOS support enabled) of the master branch at commit https://github.com/tbeu/matio/commit/17b1c34f6cc1fe6273d30386ee024edbacd600d4

### Reproduce

Take a Simulink enumeration type defined by the following code:
```
classdef(Enumeration) BasicColors < Simulink.IntEnumType
  enumeration
    Red(0)
    Yellow(1)
    Blue(2)
  end
  methods (Static = true)
    function retVal = getDefaultValue()
      retVal = BasicColors.Blue;
    end
  end
end
```

If I create an object of that type and save it to a .mat file like this (test files are attached [enum.zip](https://github.com/user-attachments/files/32267544/enum.zip)):
```
>> enumtest = [BasicColors.Blue, BasicColors.Red, BasicColors.Blue]
enumtest = 
  1×3 BasicColors enumeration array
    Blue    Red     Blue
>> save enum5.mat
```

Then a `matdump -h -d enum5.mat` shows the following (**note:** heavily edited for readability):
```
enumtest = struct {
    EnumerationInstanceTag = 3707764736
    ClassName = 1
    ValueNames = [5; 6]
    Values = CellArray [
        [3707764736, 2, 1, 1, 1, 1, 
         3707764736, 2, 1, 1, 2, 1, 
    ValueIndices = [0, 1, 0]
    BuiltinClassName = 2
    }
```

This is missing much information needed to recreate the original data.

I found that the integer values in the ClassName, ValueNames and BuiltinClassName fields are actually indices into the "strings" table in the MCOS FileWrapper__ data. The data in the "Values" array are MCOS records that can be decoded to class objects.

### Suggested fix

After applying the attached diff [enum-diff.zip](https://github.com/user-attachments/files/32267036/enum-diff.zip), the output becomes (again, heavily edited):
```
enumtest = struct {
    EnumerationInstanceTag = 3707764736
    ClassName = 'BasicColors'
    ValueNames = CellArray ['Blue'; 'Red']
    Values = CellArray [
        class BasicColors { int32.Data = 2 }
        class BasicColors { int32.Data = 0 }
        ]
    ValueIndices = [0, 1, 0]
    BuiltinClassName = 'int32.Data'
    }
```

This still is not easily readable but at least now has all information necessary to recreate the original data:
- Identifiable as an enumeration by the "EnumerationInstanceTag" and ClassName string
- ValueIndices array data are indexing into the "ValueNames" and "Values" arrays to get the original enumeration values and names:
  - 0 = Blue(2)
  - 1 = Red(1)
- BuiltinClassName gives the data type for the enumeration values.

So these changes enable an application using MatIO to decode Simulink.IntEnumType data.

However, I am not sure about the bigger impact of these changes. Certainly, writing back data that was modified like this while reading will not result in a .mat file that MATLAB can read and interpret as a proper enumeration type. On the other hand I'm not sure that would be possible with the current implementation either.

Additionally, I am concerned about my change in mat73.c, lines 650-652. It contradicts the comment in line 627 ("Groups with unknown class names (e.g. enumeration types) are not MCOS objects but struct-like containers"). With my change it is necessary to interpret the unknown class as an MCOS object, otherwise the MCOS data in the "Values" cell array does not get interpreted correctly.

Please let me know if you have further questions about this.

Best regards,
David

tbeu:
@dhansel Please test and confirm if this works for you. Thank you!

Resolves #363


dhansel:
Thanks for putting this together so quick! 

I can see two issues in my initial testing:
1. The "BuiltinClassName" in the .mat file is "int32.Data" but somehow gets shortened to "int32". That's a problem because my code uses that string to identify the proper field name in the "Values" struct array ("int32" != "int32.Data", so the actual value is not found)
2. There seems to be a problem if an enum is located within a struct or cell array. The decoding does not happen for those cases.

See attached example: [enum.zip](https://github.com/user-attachments/files/32311182/enum.zip)

dhansel:
I can confirm that with the change in [88e5240](https://github.com/tbeu/matio/pull/364/commits/88e52409e8ef3dc0edb5733e4ee9e3a96bdc0cad) the two issues I reported are resolved.

PR #366, opened 2026-09-17
Fix MCOS fully qualified class name

### LINKED ISSUE #365: MCOS class name changes when calling Mat_VarReadNext
Hi,

I have observed that for some class names (it appears for those that contain a "."), the name returned by Mat_VarGetClassName() 
is different depending on whether the variable was read by Mat_Mat_VarReadNextInfo() or Mat_VarReadNext().

For the attached file [p.zip](https://github.com/user-attachments/files/32341056/p.zip), `matdump -h p.mat` shows the class name as `Simulink.Parameter`, which is the proper name as shown in MATLAB itself.
When executing `matdump -h -d p.mat`,  the class name is shortened to just `Parameter`.

Ideally, the name should be the proper `Simulink.Parameter` in both cases.

The change appears to happen in function `ResolveMCOS` in the section of code with the comment "Prefer the active class alias...". If I deactivate that section, the name does not get shortened. I don't know for which cases this code is needed but maybe it could get changed to not prefer the alias if the alias is just a suffix of the proper name?

David

tbeu:
@dhansel Please verify and confirm once again. Thank you (also for reporting such signifcicant issues).

Resolves #365


dhansel:
Confirmed that this is working now - thanks a lot for the quick responses!

I have one final question (for now :) ) 
In the p.mat file I sent earlier, the Simulink.Parameter has a field `CoderInfo` that's showing up as `mxOPAQUE_CLASS`, i.e. it does not get resolved as an object:
```
p = Simulink.Parameter (8 fields)
    .Value = 123
    .CoderInfo: [1 mxOPAQUE_CLASS]
    .Description: [0 mxCHAR_CLASS]
    .DataType = 'auto'
    .Min: [0 mxDOUBLE_CLASS]
    .Max: [0 mxDOUBLE_CLASS]
    .DocUnits: [0 mxCHAR_CLASS]
    .Dimensions: [2 mxDOUBLE_CLASS]
```

What is special about that object so that it doesn't get resolved? This doesn't seem a general problem as objects within other objects do generally get resolved. This isn't a significant issue for me so I haven't looked into it any further but it would be interesting to know what is different here.

tbeu:
It's actually expected behavior, not a limitation of the nested resolver:

`Simulink.CoderInfo` is a **handle** object, and MATLAB stores it with a different *type system* than regular MCOS objects. In the MAT file, the opaque subelements are `type_name = "handle"` and `class_name = "Simulink.CoderInfo"` (not `type_name = "MCOS"`).

The `#subsystem#` struct has separate fields per type system (`"MCOS"`, `"handle"`, `"java"`), and matio only reads the `"MCOS"` field. `Mat_MCOS_Read5`/`Mat_MCOS_Read73` also explicitly bail out unless `type_name == "MCOS"`, so handle objects are left as `MAT_C_OPAQUE` by design.

dhansel:
Ah I see - it's not MCOS encoded - that makes sense. Thanks!

PR #371, opened 2026-09-22
Make install targets optional

Febbe:
Using matio and hdf5 as subproject has the problem, that neither hdf5 nor zlib are in a export set, but matio now expects that. This change makes that behaviour optional / controllable.


CMakeLists.txt
```diff
@@ -23,16 +23,23 @@ set(matio_PACKAGE_TARNAME "${PROJECT_NAME}")
...
-    endif()
-else()
-    if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
-        set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+# Remove this block after bumping CMake to v3.21.0
+# PROJECT_IS_TOP_LEVEL is defined then by default
+if(CMAKE_VERSION VERSION_LESS 3.21.0)
+    if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+        set(PROJECT_IS_TOP_LEVEL ON)
+        set(matio_IS_TOP_LEVEL ON)
```
MaartenBent: This new variable is not needed.

cmake/src.cmake
```diff
@@ -132,6 +132,12 @@ if(STDINT_MSVC)
 endif()
 set_target_properties(${PROJECT_NAME} PROPERTIES PUBLIC_HEADER "${MATIO_PUBLIC_HEADERS}")
 
+include(CMakeDependentOption)
+cmake_dependent_option(MATIO_INSTALL "Whether to install matio" ON PROJECT_IS_TOP_LEVEL OFF)
```
MaartenBent: I'd prefer to have all the options in `options.cmake`.

cmake/src.cmake
```diff
@@ -132,6 +132,12 @@ if(STDINT_MSVC)
...
+if(NOT MATIO_INSTALL)
+    return()
+endif()
```
MaartenBent, line 137: There is also `install(TARGETS matdump)` in `tools.cmake`. When `MATIO_INSTALL` is disabled, this should also not get installed.

MaartenBent (changes requested):
Thanks for the addition. I have a few remarks.

tbeu (changes requested):
`MATIO_INSTALL` misses documentation in README, README.md and documentation/build.texi.

MaartenBent:
Thanks. I have one more question. By using `cmake_dependent_option`, `MATIO_INSTALL` cannot be enabled at all when used in sub-project or external project.
Would it be better to use a normal `option` but change its default value?

```cmake
if(PROJECT_IS_TOP_LEVEL)
    set(MATIO_INSTALL_DEFAULT ON)
else()
    set(MATIO_INSTALL_DEFAULT OFF)
endif()
option(MATIO_INSTALL "Whether to install matio" ${MATIO_INSTALL_DEFAULT})
```

Febbe:
> Would it be better to use a normal `option` but change its default value?

Actually yes, I thought this utility allows inplace Boolean expressions. But it disables this Option when evaluated to false.
But I would use:
`option(MATIO_INSTALL "Whether to install matio" ${PROJECT_IS_TOP_LEVEL})` directly.

tbeu (approved):
LGTM

MaartenBent (approved):

